PR #4, opened 2015-04-09
Simplify logic of iterutils.chunked

dirn:
Rather than building and calculating a new list when setting a limit to
the items returned by `iterutils.chunked`, `itertools.islice` can be
used to reduce the code and take advantage of any optimizations in the
itertools module.


dirn:
Let me know if you want those commits squashed.

mahmoud:
No need to squash; I likem chunk-y ;)

PR #10, opened 2015-04-10
Various new utilities from my personal utils

ekimekim:
Some of these I find I use in every project, others are just a cool idea that I haven't ended up actually needing.
All the commits are self-contained, so please cherry-pick whichever features you think are worth having.
I'm not familiar with the markup you're using in docstrings so you may need to clean them up a bit for your liking.

For reference, this code was adapted from my "minor libraries" repo here: https://github.com/ekimekim/pylibs/
I tried to only carry over things that were general purpose and legitimately useful.

As an aside, I saw a todo that talked about a proxy object that accepted all operations. I didn't include it here, but you may want to look at my version of that. It doesn't do tracking, but could be a start: https://github.com/ekimekim/pylibs/blob/master/libs/blackhole.py

### DIFF
--- a/boltons/classutils.py
+++ b/boltons/classutils.py
@@ -1,0 +1,134 @@
+# -*- coding: utf-8 -*-
+"""Python's built-in :mod:`functools` module builds several useful
+utilities on top of Python's first-class function support.
+``classutils`` attempts to do the same for metaprogramming with
+classes, types and instances.
+"""
+
+import weakref
+
+
+def gencls(*bases, **extras):
+    """Define a new class inline (ie. class version of "lambda").
+    Takes classes to use as base class from positional arguments,
+    and any extra attributes of the class to set from keyword arguments.
+
+    Useful for once-off classes that can be described as "other class, but with a mixin"
+
+    >>> class A(object):
+    ...     name = 'Alice'
+    ...     def greet(self, x):
+    ...         print 'Hello, ' + self.name
+    ...
+    >>> class BobNameMixin(object):
+    ...     name = 'Bob'
+    ...
+    >>> gencls(A, BobNameMixin)().greet()
+    Hello, Bob
+    >>> gencls(A, name='Eve')().greet()
+    Hello, Eve
+
+    """
+    return type('gencls:{}'.format(','.join(base.__name__ for base in bases)), bases, extras)
+
+
+def issubclass(subcls, supercls):
+    """As per the builtin issubclass(), but will simply return False if
+    either arg is not suitable (eg. if subcls is not a class),
+    instead of raising TypeError.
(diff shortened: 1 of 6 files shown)


mahmoud:
OK! That's maybe the biggest contribution I've ever seen. *Lots* of interestign and good-looking stuff! It might take me a while to review. One thought, if it wouldn't be too much trouble, we could split the features up into separate pull requests and have topic-specific discussions there.

Here are my initial thoughts:
  * `partial_ordering` seems like a very good fit for funcutils. I'll test it out tomorrow.
  * `UniverseSet` also seems like solid functionality. Kind of niche, but I can see it. Could use some docs and doctests. Also, I'm curious, could you elaborate on how you've used it in the past?
  * `classproperty` is a common one that def fits. also been annoyed by the issubclass exception raising behavior.
  * `DecayCounter` is cool, and I've used this sort of thing for statistics gathering in the past. I'll need to think where it should go though.
  * `procutils` will take some time to review, looks interesting. Same goes for `termutils`. Also, `termutils` is failing CI rn.

So, in short, it may be easier to do this incrementally. `funcutils` and `setutils` changes are probably the lowest hanging fruit, if you want to make a separate PR.

Finally, is this going to be a one-time donation, or will you be down with some longer-term maintenance?

ekimekim:
Yeah, UniverseSet is the one where I thought it was cool, but haven't actually had cause to use it...
There's a bunch of silly mistakes causing build failures atm...I copy-pasted pretty haphazardly and it's possible some of this code hasn't actually been run in quite some time - trivial bugs might have crept in.
And I'm happy to stay on for the long haul. I've been wanting to explore doing a project like this for a while anyway.

Will split into more PRs either later tonight or tomorrow, not at the right computer atm.

ekimekim:
Yeah, I read the failures, they're not related to termutils at all, it was just the last commit. All trivial to fix. One was just me not knowing what the exact output would look like (`__main__.A` vs `bolton.classutils.A`).

About termutils actually, I thought I should mention: Out of all these modules, that's the one I use **alot**. In any situation where you're fiddling with `termios`, `with TermAttrs:` just saves sooo many headaches.

ekimekim:
Fixed those bugs - no time atm to split into seperate PRs but will do tomorrow.

ekimekim:
Closing - PR has been split into following smaller PRs:

https://github.com/mahmoud/boltons/pull/19
https://github.com/mahmoud/boltons/pull/18
https://github.com/mahmoud/boltons/pull/17
https://github.com/mahmoud/boltons/pull/16
https://github.com/mahmoud/boltons/pull/15
https://github.com/mahmoud/boltons/pull/14

PR #11, opened 2015-04-10
Add Python 3.4 support

rollcat:
Py2.6 compatibility was sacrificed; also 3.2 is known not to work either (mainly because of v2 unicode literals).

I've tried to keep unrelated changes away, but couldn't resist on these:

- There's `boltons.compat.lrange` that always returns a list (makes a lot of code in `statsutils.py` cleaner vs `list(xrange(x))`);
- Saw `print "Deprecation Warning: foo"` and couldn't resist converting to use the `warnings` module;
- Made `six` an install dependency; IMHO it's a better solution than rewriting all of it inside `compat.py`.

I saw a lot of random test code inside `if __name__ == "__main__"`; mostly didn't touch that, this should probably be refactored into real unit tests.


mahmoud:
This looks like top-notch stuff. I'll do a bit of review and merge it as soon as possible. I think I can probably get 2.6 compat back. It wasn't too hard with ashes, and that was doing a lot more complex stuff. Same goes for the the six dependency, since that's in line with the "each module is maximally independent" philosophy set forth in [the architecture statement](https://boltons.readthedocs.org/en/latest/architecture.html). 

Again, great stuff. I'd started on 3 compatibility last night and even though most modules were like 2-3 lines of changes, I passed out halfway through fileutils.

mahmoud:
Alright! You're merged. I did basically everything I mentioned in the comment aboove. The most affected files only got longer by 3-4 lines, and the high-degree of independence was retained. Nevertheless, Your changes were a major help, so thank you, Kamil.

PR #13, opened 2015-04-10
Functions for time since UNIX epoch

tsupinie:
I've always thought it was weird these were absent from the standard library, even though they're so simple.  I added two functions to timeutils.py to transform between seconds since the UNIX epoch and datetime objects.  Currently, they ignore any timezone information, but if you think these functions are a good fit for the package, I might be able to work on adding that kind of functionality.

### DIFF
--- a/boltons/timeutils.py
+++ b/boltons/timeutils.py
@@ -53,6 +53,7 @@ _PARSE_TD_RE = re.compile("((?P<value>%s)\s*(?P<unit>\w)\w*)" % _FLOAT_PATTERN)
 _PARSE_TD_KW_MAP = dict([(unit[0], unit + 's')
                          for _, _, unit in reversed(_BOUNDS[:-2])])
 
+_UNIX_EPOCH = datetime.datetime(1970, 1, 1, 0, 0, 0)
 
 def parse_timedelta(text):
     """Robustly parses a short text description of a time period into a
@@ -180,3 +181,26 @@ def relative_time(d, other=None, ndigits=0):
     if drt < 0:
         phrase = 'from now'
     return '%g %s %s' % (abs(drt), unit, phrase)
+
+def epoch_to_datetime(sec_epoch):
+    """Turn a number of seconds since the UNIX epoch into a datetime object.
+
+    Args:
+        sec_epoch (int or float): Number of seconds since 1 January 1970.
+
+    Returns:
+        The corresponding datetime object
+    """
+    return _UNIX_EPOCH + timedelta(seconds=float(sec_epoch))
+
+def datetime_to_epoch(dt):
+    """Get the number of seconds since the UNIX epoch corresponding to
+    a given `datetime` objects.
+
+    Args:
+        dt (datetime): A `datetime` object.
+
+    Returns:
+        Number of seconds since 1 January 1970 as a float.
+    """
+    return total_seconds(dt - _UNIX_EPOCH)
--- a/boltons/timeutils.py
+++ b/boltons/timeutils.py
@@ -185,18 +185,20 @@ def relative_time(d, other=None, ndigits=0):
 
 def epoch_to_datetime(sec_epoch):
     """Turn a number of seconds since the UNIX epoch into a datetime object.
+    Returns a naive datetime object.
 
     Args:
         sec_epoch (int or float): Number of seconds since 1 January 1970.
 
     Returns:
-        The corresponding datetime object
+        A naive datetime object that is `sec_epoch` since the UNIX epoch
     """
     return _UNIX_EPOCH + timedelta(seconds=float(sec_epoch))
 
 def datetime_to_epoch(dt):
     """Get the number of seconds since the UNIX epoch corresponding to
-    a given `datetime` objects.
+    a given `datetime` objects. Currently ignores any timezone, and assumes
+    all datetimes are UTC.
 
     Args:
         dt (datetime): A `datetime` object.
--- a/boltons/timeutils.py
+++ b/boltons/timeutils.py
@@ -16,7 +16,8 @@ from datetime import timedelta
 
 
 __all__ = ['total_seconds', 'parse_td', 'relative_time',
-           'decimal_relative_time']
+           'decimal_relative_time', 'epoch_to_datetime',
+           'datetime_to_epoch']
 
 
 def total_seconds(td):


ekimekim:
+1
Maybe be more clear about what is meant by a "naive" datetime object. I assume that means it's UTC?

mahmoud:
I guess I'm sort of wondering what the difference is between this and [`datetime.fromtimestamp()`](https://docs.python.org/2/library/datetime.html#datetime.datetime.fromtimestamp) and [`datetime.utcfromtimestamp()`](https://docs.python.org/2/library/datetime.html#datetime.datetime.utcfromtimestamp)

tsupinie:
@mahmoud Haha, there isn't any. How did I not see those?  Carry on ...

ekimekim:
I don't see anything in the stdlib for datetime to epoch though? Only datetime to time tuple...

mahmoud:
That's a good point. A datetime to timestamp is pretty trivial right? In 2.7+, `(dt - epoch_dt).total_seconds()`. Still, might be a useful thing for `timeutils` for 2.6 folks.

markrwilliams:
Firstly, it's difficulty to correctly convert between UTC time and Epoch time because [Epoch time does not count leap seconds](http://en.wikipedia.org/wiki/Unix_time).  Consequently it's also difficult to convert between local times and Epoch time.

I still think we should have a function that converts between Epoch time and datetimes because Python already provides `*fromtimestamp` methods on datetime.  But we should be sure to document this difficulty.

Unfortunately the pull request's `datetime_to_epoch` implementation can't convert a timezone aware `datetime` to Epoch time at all!  That's because Python disallows arithmetic between timezone aware and naive `datetime` objects.

It's possible that we can use the current implementation in the event that the datetime is naive, and just assume that it's UTC.  That seems wrong to me, though.

With that in mind I did some  very basic research.  It seems that the best solution involves [`time.mktime`](https://docs.python.org/2/library/time.html#time.mktime). 

It's the inverse of [`time.locatime`](https://docs.python.org/2/library/time.html#time.localtime), though, so it assumes that the incoming `time.struct_time` object represents a moment in the same timezone as the program in which it's called.  

Consequently our `datetime_to_epoch` function should require the `datetime` object be timezone aware, so that it can convert it to local time via [`datetime.datetime.astimezone`](https://docs.python.org/2/library/datetime.html#datetime.datetime.astimezone).  It also needs the system's local timezone.

Below is a sketch that shows one way to put these ideas together:

````python
from boltons.tzutils import Pacific, Eastern, UTC
import time


def datetime_to_epoch(dt, timezone):
    if not dt.tzinfo:
        raise ValueError('datetime_to_epoch'
                         ' cannot be applied to a naive datetime')
    if dt.tzinfo != timezone:
        dt = dt.astimezone(timezone)

    return time.mktime(dt.timetuple())
````

And here's what it looks like in action:

````python
import datetime
now = datetime.datetime.now().replace(tzinfo=Pacific)
print datetime_to_epoch(now, timezone=Pacific)
print datetime_to_epoch(now, timezone=Eastern)
print datetime_to_epoch(now, timezone=UTC)
````

A nicer API would not require the `timezone` argument, but instead would use a new timezone in boltons along the lines of [dateutil.tz.tzlocal](https://labix.org/python-dateutil#head-5fb12f4538c5a2fd83f87eea8e6c0ddd47f8b4b0).

**Comment by mahmoud** [comment]
Umm, you mean like https://boltons.readthedocs.org/en/latest/tzutils.html#boltons.tzutils.LocalTZ ?

**Comment by markrwilliams** [comment]
Sure!  This is what I get for trying to finish a comment before I go out the door.

Now, what about the rest of what I said?

**Comment by mahmoud** [comment]
Well, I think it's very correct. Practically speaking, however, for most use cases people are probably just looking for a "datetime to seconds-based timestamp". It may not be the bestest practice, but I use naive datetimes all the time when whipping up quick scripts. So this has me thinking:

  1. Allow naïve datetimes and just do the last line
  2. Maybe ``tzutils`` should move into ``timeutils``. ``tzutils`` already wants ``total_seconds`` and ``timeutils`` will then want LocalTZ.

And we'll just document all of this, linking people to ``dateutil`` if they need more advanced behavior. Thoughts?

**Comment by markrwilliams** [comment]
Re 1: You've got a super good point about the use of naive datetime objects. Perhaps unsurprisingly, `datetime.datetime.now()` returns whatever your time your system thinks "now" is -- which is the definition of local time, after all! And if it's safe to assume a naive `datetime` is local time, then it's also safe to pass `time.mktime()` a naive datetime.

Now what if there is a timezone?  With `LocalTZ` we can easily convert it, inverting the previous logic like so:

````python
def datetime_to_epoch(dt):
    if dt.tzinfo:
        dt = dt.astimezone(LocalTZ)

    return time.mktime(dt.timetuple())
````

(If `dt.tzinfo` is the same as `LocalTZ` the conversion is a noop.)

Re 2: I think moving `tzutils` into `timeutils` will save me further embarassment.  So it's a great idea!

**Comment by zed** [comment]
@tsupinie there is a tiny difference e.g., `datetime.utcfromtimestamp()` expects a non-POSIX timestamp on some systems if the local timezone is "right" e.g., `TZ=right/UTC`. While `posix_epoch + timedelta(seconds=posix_timestamp)` always produces UTC time (and the expression is plain wrong if you give it the result of `time.time()` when it returns non-POSIX timestamp). Both can be valid depending on your input and the expected output. See [discussion in the comments](http://stackoverflow.com/a/28950133/4279). 

`timeutils.total_seconds()` should be tested whether it has [floating point issues (or you could just use the same formula as Python 2.7)][1].

@markrwilliams `datetime` module can't represent a positive leap second therefore there are no issues with `datetime` and POSIX timestamp:

    from datetime import datetime, timedelta, timezone

    utc_naive = datetime(1970, 1, 1) + timedelta(seconds=posix_timestamp)
    utc_aware = datetime(1970, 1, 1, tzinfo=timezone.utc) + timedelta(seconds=posix_timestamp)

where `timezone.utc` is not available on Python 2 but [it is easy to define it there](http://stackoverflow.com/a/25421145/4279).

UTC time "forgets" leap seconds i.e., `(utc_then - utc_now)` doesn't represent the true elapsed time (it doesn't count leap seconds in between).

@markrwilliams your `datetime_to_epoch()` implementation may fail unnecessary during end-of-DST transitions when the local time is ambiguous e.g.,  `2015-11-01 01:30:00` in America/New_York timezone and (more complex case) `2014-10-26 01:30:00` in Europe/Moscow timezone. 

To convert a naive datetime object that represents time in the local timezone in such cases, you need to [specify `is_dst` parameter to resolve ambiguity](https://mail.python.org/pipermail/python-dev/2015-April/139167.html):

```
from datetime import datetime
import pytz # $ pip install pytz

tz = pytz.timezone('Europe/Kiev')
print(tz.localize(datetime(1990, 7, 1, 1, 30), is_dst=False).strftime('%c %z %Z'))
# -> Sun Jul  1 01:30:00 1990 +0300 EEST
print(tz.localize(datetime(1990, 7, 1, 1, 30), is_dst=True).strftime('%c %z %Z'))
# -> Sun Jul  1 01:30:00 1990 +0400 MSD
```

`is_dst` might not be enough if the ambiguous local time may occur more than twice (I don't know whether there is such timezone) but it should be "good enough" in most cases.

It is *always* possible to convert an aware datetime object (any timezone) or a naive datetime object that represents time in UTC to POSIX timestamp unambiguously, see [Converting datetime.date to UTC timestamp in Python][1]. 

To get the system local timezone, don't use ` dateutil.tz.tzlocal` or `timeutils.LocalTZ`, [use `tzlocal.get_localzone()` instead](http://stackoverflow.com/a/17365806/4279).

Both `pytz` and `tzlocal` are pure Python. [They (or equivalent) might be included in stdlib eventually](https://www.python.org/dev/peps/pep-0431/). `pytz` license is MIT (author says: *"I’m happy to relicense this code if necessary for inclusion in other open source projects."*). `pytz` might be too low-level to expose it to the user directly; [`times` package](https://pypi.python.org/pypi/times/0.7) provides a nice simple interface but it is implemented in terms of `arrow` library that had issues with timezone conversions last time I've checked and [those issues might not be fixed](https://github.com/crsmithdev/arrow/issues/4).

[1]: http://stackoverflow.com/a/8778548/4279

**Comment by mahmoud** [comment]
Hey @zed/JFS, that's super informative, thanks! I was suspicious of datetime and leapseconds, but I'm hardly an expert. That said, I think @markrwilliams might have been referring to the system's disposition toward leapseconds, not Python's. I'll let him jump in if that's the case.

The `boltons` stance toward time support is that past a certain point, time is as hard as you want it to be. The docs should give a lot of pointers forward to third-party libraries that make for the most correct system, if that should end up being the goal of the application developer. Awareness gradient, etc. I mention this because I hadn't heard about `tzlocal`. I'll definitely link to it in the boltons docs, as well as some of the other methods described here.

Judging by the conversation, it's probably best to go the simpler route of just making this work with UTC. boltons [defines a small `tzinfo` for UTC](https://boltons.readthedocs.org/en/latest/tzutils.html#boltons.tzutils.UTC), as an instance of [ConstantTZInfo](https://github.com/mahmoud/boltons/blob/master/boltons/tzutils.py#L45). Unless of course, @zed, you have some interest in submitting a pull request representing the techniques described in your encyclopedic Stack Overflow answer.

**Comment by zed** [comment]
@mahmoud Good point on "Awareness gradient". Expect brokenness when you deal with timezones. Though for many applications there are "good enough" solutions.

`pytz` provides access to [the tz database](https://www.iana.org/time-zones/repository/tz-link.html):

> This database (often called zoneinfo or tz) is used by several implementations, including the GNU C Library (used in GNU/Linux), Android, Firefox OS, FreeBSD, NetBSD, OpenBSD, Cygwin, DJGPP, MINIX, webOS, AIX, BlackBerry 10, iOS, Microsoft Windows, OpenVMS, Oracle Database, Oracle Solaris, and OS X.

There are known issues with `dateutil.tz.tzlocal` and  `time` module -​based solutions such as `tzutils.LocalTZ`. Unless you have a very good reason, you should use `pytz` for the timezone-related code (perhaps indirectly via a 3rd-party library). 

All one need to know about `tzlocal` is that `tzlocal.get_localzone()` returns the local timezone as `pytz` timezone. It is not perfect but it is "good enough" in many cases.

`datetime.fromtimestamp()` and `datetime.timestamp()` (or its backport) methods  are acceptable in many cases especially if you provide the corresponding pytz timezone. 

---
<sup>you can skip the rest</sup>

The point about leap seconds is that they do not affect UTC <-> [seconds since the Epoch][1] conversion unless you need to [convert the moment during a leap second such as `2015-06-30 23:59:60+0000`](http://stackoverflow.com/a/21029510/4279) i.e., *every* UTC time that can be represented as a `datetime` object can be converted to POSIX time and back e.g., nobody knows how many leap seconds will occur between now and `January 1, 2100` but it is easy to find the corresponding POSIX timestamp:

```
>>> from datetime import date, timedelta
>>> (date(2100, 1, 1) - date(1970, 1, 1)) // timedelta(seconds=1)
4102444800
```

It is the *exact* result ([by definition](https://gist.github.com/zed/ff4e35df3887c1f82002)). And in reverse:

```
>>> datetime(1970, 1, 1) + timedelta(seconds=2147483647)
datetime.datetime(2038, 1, 19, 3, 14, 7)
```

`2038-01-19 03:14:07+0000` is exact result no matter how many leap seconds in between. See [Seconds Since the Epoch][1]. You could consider POSIX time as a number of non-leap SI seconds since the Epoch (`1970-01-01 00:00:00+0000`) or the number of elapsed UT1 (mean solar) seconds, see [Is a day always 86,400 epoch seconds long?](http://stackoverflow.com/a/20335352/4279).

---

The current "computer time" can be related to UTC time in different ways: POSIX, Mills, smear, ntp, TAI. 

[1]: http://pubs.opengroup.org/stage7tc1/basedefs/V1_chap04.html#tag_04_15

**Comment by mahmoud** [comment]
OK resurrecting ancient thread. I read and reread every comment twice. Unless I missed something I think I'm clear to do the following. For boltons 15.0 I'm doing a `dt_to_timestamp` that will be based on on the subtraction method I described in the 6th post of this thread. It sidesteps all the timezone complexities and keeps the behavior pretty predictable and suitable for boltons. 

tzutils will merge into timeutils and I will strive to document all the rest of this as usefully as possible in the docs.

**Comment by mahmoud** [comment]
Also, @zed , I was unable to stimulate any of the error conditions in floating point precision. I added the timedelta.total_seconds tests per the issue's diff here: 1489864cd22db0b1a3fbd793805639cd406d1b8f . All of them passed without any changes. I really really don't know why that Python issue got created. Future/true division is needless overkill.

**Comment by mahmoud** [comment]
And like that, after many months of dormancy, I think that this pull request is now resolved. @tsupinie thank you for the suggestion. I hope that the new dt_to_timestamp function coming out in 15.0's new and improved timeutils will suit your needs nicely. Thanks again everyone!

**Comment by zed** [comment]
@mahmoud: I can reproduce [the floating point issue](http://bugs.python.org/issue8644):

```
>>> from __future__ import division
>>> (2**53+1)/10**6
9007199254.740993
>>> (2**53+1)/1e6
9007199254.740992
```

Note: the results of `/10**6` (with truediv) and `/1e6` differ.

mahmoud:
Try doing it with a timedelta. timeutils.total_seconds() and datetime.total_seconds() get the same results for a wide range of test cases, including all those from the standard library. A few have been added to /tests/test_timeutils.py

PR #14, opened 2015-04-11
timeutils: Add a DecayCounter which decays in value with time

ekimekim:
Since DecayCounter can fail badly when subject to time jumps, it makes an attempt to use
monotonic time when available.

Though I initially just wrote this up because it sounded cool, I've since used it to implement a rate-limit counter. It has interesting properties for this use-case since it is much more sensitive to spikes than constant near-limit load. The original use was intended to be for stats.

### DIFF
--- a/boltons/timeutils.py
+++ b/boltons/timeutils.py
@@ -21,23 +21,27 @@ try:
     # python3.3 added monotonic time to stdlib
     monotonic = time.monotonic
 except AttributeError:
-    from importlib import import_module
-    # lib name, function that gets value given module (in order of preference)
-    _monotonic_libs = [
-        ('monotonic', lambda module: module.monotonic()),
-        ('monotime', lambda module: module.monotonic()),
-        ('monoclock', lambda module: module.nano_count() / 1e9),
-    ]
-    for _lib, _func in _monotonic_libs:
-        try:
-            _monotonic_module = import_module(_lib)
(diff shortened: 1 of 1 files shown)


boltons/timeutils.py
```diff
@@ -11,12 +11,41 @@
...
+    # python3.3 added monotonic time to stdlib
+    monotonic = time.monotonic
+except AttributeError:
+    try:
+        from importlib import import_module
+    except ImportError:
+        monotonic = None # py2.6
+    else:
+        # lib name, function that gets value given module (in order of preference)
+        _monotonic_libs = [
```
markrwilliams: It would be better to explicitly pass in a `monotonic` function to the `DecayCounter` instance.  Not only is the logic clearer, it also enables me to use arbitrary bindings to my platform's appropriate monotonic time interface without monkeypatching the module

ekimekim: That's a good point.
I might keep the >= python3.3 `time.monotonic()` check though, and have it always use that if available.
Then python3 and python2 users both get a sane, expected behaviour.

mahmoud:
Closing, partially due to the monotonic troubles, partially due to a sense that this is more a statistical counter than a time utiltiy, but mostly due to [this logic](https://github.com/mahmoud/boltons/pull/17#issuecomment-114378322). Thank you for the contribution (and subsequent patience)!

PR #15, opened 2015-04-11
classutils: New module with utils based on classes, types and instances

ekimekim:
Introduces a variety of metaprogramming helpers to do with classes.
It's possible some of the things already in funcutils (eg. mro_items()) would more properly belong here.

I only rarely find use for gencls (it's mainly helpful when dealing with multiple inheritance), but everything else here I get regular use out of. For example, I use `TracksInstances` in combination with `get_all_subclasses()` to implement plugin systems where all instances of subclasses of the Plugin class are automatically found and used.

### DIFF
--- a/boltons/classutils.py
+++ b/boltons/classutils.py
@@ -1,0 +1,134 @@
+# -*- coding: utf-8 -*-
+"""Python's built-in :mod:`functools` module builds several useful
+utilities on top of Python's first-class function support.
+``classutils`` attempts to do the same for metaprogramming with
+classes, types and instances.
+"""
+
+import weakref
+
+
+def gencls(*bases, **extras):
+    """Define a new class inline (ie. class version of "lambda").
+    Takes classes to use as base class from positional arguments,
+    and any extra attributes of the class to set from keyword arguments.
+
+    Useful for once-off classes that can be described as "other class, but with a mixin"
+
+    >>> class A(object):
+    ...     name = 'Alice'
+    ...     def greet(self):
+    ...         print 'Hello, ' + self.name
+    ...
+    >>> class BobNameMixin(object):
+    ...     name = 'Bob'
+    ...
+    >>> gencls(BobNameMixin, A)().greet()
+    Hello, Bob
+    >>> gencls(A, name='Eve')().greet()
+    Hello, Eve
+
+    """
+    return type('gencls:{}'.format(','.join(base.__name__ for base in bases)), bases, extras)
+
+
+def issubclass(subcls, supercls):
+    """As per the builtin issubclass(), but will simply return False if
+    either arg is not suitable (eg. if subcls is not a class),
+    instead of raising TypeError.
+    """
+    try:
+        return __builtins__['issubclass'](subcls, supercls)
+    except TypeError:
+        return False
+
+
+def get_all_subclasses(cls):
+    """Recursive method to find all decendents of cls, ie.
+    its subclasses, subclasses of those classes, etc.
+    Returns a set.
+
+    >>> class A(object):
+    ...     pass
+    ...
+    >>> class B(A):
+    ...     pass
+    ... 
+    >>> class C(B):
+    ...     pass
+    ... 
+    >>> class D(A):
+    ...     pass
+    ... 
+    >>> get_all_subclasses(A)
+    set([<class 'boltons.classutils.D'>, <class 'boltons.classutils.B'>, <class 'boltons.classutils.C'>])
+    >>> get_all_subclasses(B)
+    set([<class 'boltons.classutils.C'>])
+
+    """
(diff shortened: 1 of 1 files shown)


ekimekim:
```
Expected:
    set([<class 'boltons.classutils.D'>, <class 'boltons.classutils.B'>, <class 'boltons.classutils.C'>])
Got:
    set([<class 'boltons.classutils.C'>, <class 'boltons.classutils.B'>, <class 'boltons.classutils.D'>])
```
any ideas on how to make these doctests less fragile?

textbook:
See e.g. http://stackoverflow.com/q/15549429/3001761

ekimekim:
Good tips there. I went with expressing the return value by `thing() == value` - I think it's actually more readable than before, now.

ekimekim:
Sigh, tabs. Forgot that I hadn't turned on expandtab for this project yet. Hopefully won't happen again.
(I use tabs in my personal projects, and selectively enable expandtab on projects that don't use them)

ekimekim:
Finally. 2.6 support is really annoying.

mahmoud:
Finally getting to the backlog here. I'm going to be putting some comments inline.

boltons/classutils.py
```diff
@@ -0,0 +1,132 @@
...
+"""Python's built-in :mod:`functools` module builds several useful
+utilities on top of Python's first-class function support.
+``classutils`` attempts to do the same for metaprogramming with
+classes, types and instances.
+"""
+
+import weakref
+
+
+def gencls(*bases, **extras):
```
mahmoud: I'm skeptical that this one-liner is a common enough trope among python programmers to warrant inclusion. It's just such a thin wrapper around type that I'd expect the standard code review response would be explicit better than implicit: spell out the class.

boltons/classutils.py
```diff
@@ -0,0 +1,132 @@
...
+
+    def __init__(self, fn):
+        self.fn = fn
+
+    def __get__(self, instance, cls):
+        arg = cls if instance is None else instance
+        return lambda *args, **kwargs: self.fn(arg, *args, **kwargs)
+
+
+class TracksInstances(object):
```
mahmoud: I've done this sort of thing in projects before, and while I found it useful at the time, I still consider it too fancy (esp with the weakrefs) to be the sort of pattern I would actively evangelize. It is fun though.

boltons/classutils.py
```diff
@@ -0,0 +1,132 @@
...
+    >>> get_all_subclasses(B) == set([C])
+    True
+
+    """
+    subs = set(cls.__subclasses__())
+    subs_of_subs = [get_all_subclasses(subcls) for subcls in subs]
+    return subs.union(*subs_of_subs)
+
+
+class classproperty(object):
```
mahmoud: This one's an old standby. Lots of folks ask about it and this is a pretty standard approach.

boltons/classutils.py
```diff
@@ -0,0 +1,132 @@
...
+    For simplicity, only read-only properties are implemented."""
+
+    def __init__(self, fn):
+        self.fn = fn
+
+    def __get__(self, instance, cls):
+        return self.fn(cls)
+
+
+class mixedmethod(object):
```
mahmoud: In my opinion this is definitely an antipattern. I would strongly encourage this sort of thing (which is hopefully rare) to be implemented with a `staticmethod` with explicit disambiguation done within the method itself.

boltons/classutils.py
```diff
@@ -0,0 +1,132 @@
...
+    >>> gencls(BobNameMixin, A)().greet()
+    Hello, Bob
+    >>> gencls(A, name='Eve')().greet()
+    Hello, Eve
+
+    """
+    return type('gencls:{0}'.format(','.join(base.__name__ for base in bases)), bases, extras)
+
+
+def issubclass(subcls, supercls):
```
mahmoud: Definitely run into this before. :+1:

boltons/classutils.py
```diff
@@ -0,0 +1,132 @@
...
+    either arg is not suitable (eg. if subcls is not a class),
+    instead of raising TypeError.
+    """
+    try:
+        return __builtins__['issubclass'](subcls, supercls)
+    except TypeError:
+        return False
+
+
+def get_all_subclasses(cls):
```
mahmoud: Simple and explicit enough in purpose, but needs to be made recursive to be correct. I can whip that up though.

mahmoud:
So, I'm going to merge this in as a new module, making it the first community-contributed bolton module :smile:. Note that it will probably end up being named `typeutils`, since this is modern Python after all :wink:

boltons/classutils.py
ekimekim: It already is recursive. See line: `subs_of_subs = [get_all_subclasses(subcls) for subcls in subs]`

PR #16, opened 2015-04-11
add new file procutils, for process-related utilities

ekimekim:
Add a more flexible interface to subprocess, in the style of subprocess.call() but better.

I find `subprocess.call()`, `check_call()` and `check_output()` really badly lacking, but I end up using them anyway due to the considerable extra complexity in using my own `Popen` object.

For example:
* `check_*` does not report an exit code or stderr in the exception, so you may know "something failed" but it's hard to tell what or why.
* `*_call` use the process's stdout, resulting in often unwanted extra output
* with any of these calls, it is impossible to interact with stdin

I have endeavoured to create a function that is very flexible but with excellent defaults, in the style of `requests.get()`, where the 90% use case is very simple, but the 99% use case is still possible. In getting into specifics (such as having None redirect output to `/dev/null`), I'm pretty sure I broke windows support. It's probably possible to add with minor changes in interface though.

### DIFF
--- a/boltons/procutils.py
+++ b/boltons/procutils.py
@@ -29,7 +29,7 @@ def _raise(ex_type, ex, tb):
 
 def cmd(args, return_stdout=True, return_stderr=False, return_code=False, return_proc=False,
         stdin=None, stdout=PIPE, stderr=PIPE, success=0, cwd=None, close_fds=True, env={}, clear_env=False):
-    """Run a command, blocking until it exits.
+    r"""Run a command, blocking until it exits.
     By default, returns the command's stdout as a string.
     First argument "args" should be either a list of arguments,
     or a string (which will be subject to shell interpretation).
--- a/boltons/procutils.py
+++ b/boltons/procutils.py
@@ -43,7 +43,7 @@ def cmd(args, return_stdout=True, return_stderr=False, return_code=False, return
     Note that this function only supports POSIX platforms.
 
     Simple usage:
-        >>> cmd(['echo', 'hello world'])
+        >>> cmd(['echo', 'hello world']) # doctest: +SKIP
         'hello world\n'
 
     A note on return values:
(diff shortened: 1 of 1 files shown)


ekimekim:
`ValueError: line 14 of the docstring for boltons.procutils.cmd has inconsistent leading whitespace: "'"`
I don't understand what's wrong with the doctest here?

```
E     File "/home/travis/build/mahmoud/boltons/boltons/procutils.py", line 143
E       raise ex_type, ex, tb
E                    ^
E   SyntaxError: invalid syntax
```
sigh, some kind of py3 compat issue. I'll need to look up how to raise exceptions without clobbering tracebacks in the new system.

textbook:
See e.g. http://stackoverflow.com/q/15838224/3001761

ekimekim:
Right, seems it's actually simpler in py3 than py2 (for those wondering, the purpose behind explicitly raising a certain tb instead of just `raise` with no args is that the inner try block may clobber the traceback from the outer except block).

What's harder is that this is a **Syntax**Error. Which means a typical:
```python
if IS_PY3:
    raise ex
else:
    raise ex_type, ex, tb
```
isn't going to work. How do people normally solve this?

textbook:
Looks like this would require either `six` or `future` - see e.g. http://python-future.org/compatible_idioms.pdf

mahmoud:
Note that the six requirement has been dropped. That said, just check inside six to see how it does what you need, it'll probably be shorter than you expect.

ekimekim:
Found it but it's nasty. Wraps the code in `exec()` so it isn't parsed until runtime. Will update PR.

ekimekim:
Ok, this is the only problem left:
`ValueError: line 14 of the docstring for boltons.procutils.cmd has inconsistent leading whitespace: "'"`
The offending code:
```python
"""<snip>
    Simple usage:
        >>> cmd(['echo', 'hello world'])
        'hello world\n'

    A note on return values:
<snip>"""
```
...anyone have any idea what's going wrong here?

asfaltboy:
I think you should escape the newline `\n` char or make the whole dosctring raw (`r""" ... """`), as seen in [this SO question](http://stackoverflow.com/questions/18772991).

ekimekim:
Of course, i'm an idiot. Didn't see that,

mahmoud:
Closing, per https://github.com/mahmoud/boltons/pull/17#issuecomment-114378322

PR #17, opened 2015-04-11
termutils: New file for termials/termios-related utilities

ekimekim:
I use my computer mainly in a terminal environment, and often find myself needing to do tasks to do with terminal input/output. For example, wanting to turn off echoing and accept input one-char-at-a-time, to implement key-shortcut-controlled interfaces (like htop, mplayer).
Other times, I find myself wanting to know the width/height of the screen for the purposes of displaying output (the COLUMNS env var is not always reliable).

The `TermAttrs` class greatly streamlines access to the `termios.tc{get,set}attrs` calls. You always want to reset your termattrs on exit, or else you mess up the terminal for the user's shell. This involves, every time:
```
old_attrs = get_attrs()
try:
    set_attrs(new_attrs)
    stuff
finally:
    set_attrs(old_attrs)
```
With `TermAttrs`, this becomes:
```
with TermAttrs(new_attrs):
    stuff
```
indeed, this is a textbook example of where a context manager is useful, and it surprised me the stdlib did not implement one.

`termsize()` caches results and uses SIGWINCH to update. This may be considered by some as "too invasive" for a utility library (though I would like to point out that `subprocess` installs a SIGCHLD handler which it requires in order to work correctly) - maybe consider just doing the ioctl every time? Overhead shouldn't be that high, but I kinda designed the function to be called on every "frame" of a tight looping interface.

### DIFF
--- a/boltons/termutils.py
+++ b/boltons/termutils.py
@@ -1,0 +1,93 @@
+
+"""Utilities for working with terminals.
+This supplements the stdlib module `termios`, and is only available for posix platforms.
+"""
+
+import fcntl
+import signal
+import struct
+import sys
+import termios
+
+__ALL__ = ['TermAttrs', 'termsize']
+
+
+class TermAttrs(object):
+    def __init__(self, attrs, fd=None, when=termios.TCSANOW):
+        """A context manager for changing terminal attrs.
+        fd defaults to stdin.
+        attrs must be a 7-element iterable as taken by tcsetattr.
+
+        Note: context manager is not reenterant wrt the same instance.
+        """
+        self.attrs = attrs
+        self.fd = fd if fd is not None else sys.stdin.fileno()
+        self.when = when
+
+    @classmethod
+    def modify(cls, include=(0,0,0,0), exclude=(0,0,0,0), fd=None, *args, **kwargs):
+        """Alternate creation function, allowing you to base changes off current term attrs.
+        include and exclude should be 4-tuples of (iflag, oflag, cflag, lflag).
+        All values in include will be set.
+        All values in exclude will be unset.
+        All other values will be unchanged from current values.
+        Other args are passed though to TermAttrs.__init__
+        """
+        if fd is None: fd = sys.stdin.fileno()
+        attrs = termios.tcgetattr(fd)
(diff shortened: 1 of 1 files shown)


boltons/termutils.py
```diff
@@ -0,0 +1,93 @@
...
+        global _size_cache
+        _size_cache = columns, rows
+
+    def termsize_handler(frame, signal):
+        get_termsize()
+        if old_handler is not None and callable(old_handler):
+            old_handler()
+
+    if _size_cache is None:
+        old_handler = signal.signal(signal.SIGWINCH, termsize_handler)
```
markrwilliams: Unfortunately you can't safely install a signal handler in Pythons prior to 3.5!  Most signals in Python retain their default disposition, which mostly (again) is to terminate the program (see man 7 signal) or to ignore the signal.

Consequently very little Python code that makes system calls is prepared to handle EINTR. When you set your own handler for some signal, that signal will now cause system calls to terminate with an EINTR error. These two facts together mean that registering signal handlers in Python is serious business!  

For example, using this code will mean that if you have a `time.sleep` or a `socket.recv` in your main thread, resizing the terminal will cause that call to fail with EINTR!

Python 3.5 attempts to resolve this: https://www.python.org/dev/peps/pep-0475/  In prior versions it's better to either not set signal handlers at all, or to only set them as part of some larger event handling system (as Twisted does).

ekimekim: I see. I guess I've never come across this problem in practice because I'm more often using `gevent` which, as you mentioned, has its own signal handling mechanism that doesn't cause random EINTRs thoughout the code.

Best to just remove the signal handler and caching logic entirely then.

mahmoud:
The time has come to make a decision on these open pull requests. I'm preparing boltons for its first major release.  After much deliberation and revisions to the [architecture statements](http://boltons.readthedocs.org/en/latest/architecture.html#design-of-a-bolton), I'm going to be taking a pass on anything complex that I haven't used/needed myself. As soon as I do something terminal-related, you can bet I'll be revisiting this. In the meantime, thank you for your patience and contributions, looking forward to next time!

PR #18, opened 2015-04-11
setutils: Add UniverseSet(), a virtual set containing "all" items

ekimekim:
Ok, I admit this one's a bit esoteric. A `UniverseSet()` contains **literally every object possible** except ones you've specifically excluded. It's compatible with all set operations, eg:
```python
whitelist = UniverseSet()
whitelist -= blacklist
for value in whitelist & values:
  do_a_thing(value)
```

I haven't really had a good use for it (eg. the above could be written much simpler without it), but I thought it was really cool.
Maybe a more general "infinite set" where membership can be filtered by functions (eg. `PredicateSet(lambda x: int(x) == x and x >= 0)` would be the set of natural numbers)? Or maybe that should remain a more niche mathematics library thing?

What do you think? Is this possibly useful enough to consider including?

...actually now that I've written this, I think I'd prefer to not merge this. But I'll leave it up to you.

### DIFF
--- a/boltons/setutils.py
+++ b/boltons/setutils.py
@@ -8,6 +8,8 @@ of this, sets are not indexable, i.e, ``my_set[8]`` will raise an
 :exc:`TypeError`. The :class:`IndexedSet` type remedies both of these
 issues without compromising on the excellent complexity
 characteristics of Python's built-in set implementation.
+The :class:`UniverseSet` type extends python's finite set implementation to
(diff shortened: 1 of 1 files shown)


mahmoud:
After long (maybe too long) consideration, I'm going to err on the side of conservatism and not merge this. Thanks for your hard work and patience nonetheless. I swear the first time I see an application, I'm coming straight back here to reverse my decision. :)

PR #26, opened 2015-04-12
poolutils: Pool implementations

ekimekim:
A generic pool. It handles arbitrary "resources", which could be files, connections, etc etc. Supports "cleaning" resources in between uses (eg. postgresql connections require this behaviour) as well as bursting under high load then destroying excess resources back to a limit later. Reasonable defaults thoughout.

Currently PR contains only the generic pool, but specific subclasses (threadpool, socketpool...) to follow.

### DIFF
--- a/boltons/poolutils.py
+++ b/boltons/poolutils.py
@@ -240,7 +240,7 @@ class ResourceWrapper(object):
         return getattr(self._resource, attr)
 
     def __setattr__(self, attr, value):
-        if not hasattr(self, '_resource'):
+        if '_resource' not in self.__dict__:
             return super(ResourceWrapper, self).__setattr__(attr, value)
         return setattr(self._resource, attr, value)
 
--- a/boltons/poolutils.py
+++ b/boltons/poolutils.py
@@ -1,0 +1,243 @@
+
+"""Implementations of various kinds of resource pool"""
+
+__all__ = ['Pool']
+
+
+class Pool(object):
+    """Implements a generic pool of resources.
+    The resource is an object that:
+    * is created by the given factory function / class
+    * is intended to be re-used
+    * is hashable and unique (ie. multiple calls to factory() won't return the same object)
+    * optionally, gets prepared for re-use by given clean() function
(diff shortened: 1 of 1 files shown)


ekimekim:
Ok, so it seems I never actually ran this code. Still lots of bugs. Will need to work on it properly later.

doublereedkurt:
select.select() can also be used to check if sockets are in an error state.  (If your design is to check sockets one at a time, there isn't any particular advantage however.)

It is generally good practice to try to shutdown() sockets before closing (this allows the server to free resources faster.)
socket.close() may raise exceptions.  Especially if it is an SSL socket, they will often do an SSL protocol shutdown handshake on close.

Which also brings up that close may timeout or take a long time :-)  (It will call SSL_shutdown https://www.openssl.org/docs/ssl/SSL_shutdown.html) 

This stuff gets VERY tricky.  I've not attempted to write code to solve it in the general case, only for a known model of concurrency (gevent), which I am also able to control all of the "socket-like" things that go in.

mahmoud:
Yeah, I'm going to have to agree with Kurt. Connection caching and indeed pooling of any limited resource is very, *very* dependent on the semantics of that resource. On the surface it seems like a commonly-necessitated feature: SQLAlchemy, requests, paramiko, SuPPort, and many others all use some form of pooling. But if one takes a real hard look at how libraries use these things, it varies widely. It was a challenge to find two libraries or applications where this would be a suitable drop-in strategy for both. Not that they don't exist, and not that this isn't a much more robust solution than ones I've used in the past, but I think it may exert too much influence on the architecture/strategy in an application's design.

And that brings me to the crux of the issue. `boltons` isn't a library of strategies. It's about more robust tactics and saving a few hastily hacked-together lines from causing bugs. It's nothing overarching like a concurrency or pooling strategy. With a name like `Pool`, there's not really a bright line on which concurrency strategies and protocols are supported. A generic pool is a cache, and the ones in cacheutils are about as complex as boltons was meant to get.

PR #29, opened 2015-04-18
New module for math utilities. Includes alternative ceiling/floor functions.

blebo:
New module for math utilities. Includes alternative ceiling/floor functions.

### DIFF
--- a/boltons/sockutils.py
+++ b/boltons/sockutils.py
@@ -1,0 +1,317 @@
+# -*- coding: utf-8 -*-
+
+import time
+import socket
+
+
+try:
+    from typeutils import make_sentinel
+    _UNSET = make_sentinel(var_name='_MISSING')
+except ImportError:
+    _UNSET = object()
+
+DEFAULT_TIMEOUT = 10  # 10 seconds
+DEFAULT_MAXBYTES = 32 * 1024  # 32kb
+
+
+class BufferedSocket(object):
+    def __init__(self, sock,
+                 timeout=DEFAULT_TIMEOUT, maxbytes=DEFAULT_MAXBYTES):
+        self.sock = sock
+        self.sock.settimeout(None)
+        self.rbuf = b''
+        self.sbuf = []
+        self.timeout = timeout
+        self.maxbytes = maxbytes
+
+    def settimeout(self, timeout):
+        self.timeout = timeout
+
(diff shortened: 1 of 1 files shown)


boltons/mathutils.py
```diff
@@ -0,0 +1,44 @@
+"""This module provides useful math functions on top of Python's built-in :mod:`math` module.
+"""
+
+
+def ceil_from_list(lst, x, allow_equal=True):
```
mgaitan: why `_from_list` ? it actually allows any iterable   .  And  that probably means it would better fit in the `itertools` module.

boltons/mathutils.py
```diff
@@ -0,0 +1,44 @@
...
+    >>> VALID_CABLE_CSA = [1.5, 2.5, 4, 6, 10, 25, 35, 50]
+    >>> ceil_from_list(VALID_CABLE_CSA, 3.5)
+    4
+    >>> ceil_from_list(VALID_CABLE_CSA, 4)
+    4
+    >>> ceil_from_list(VALID_CABLE_CSA, 4, allow_equal=False)
+    6
+    """
+
+    return min(filter(lambda y: y >= x if allow_equal else y > x, lst))
```
mgaitan: btw, using `min()` is far of being the optimal implementation.  you should check this answer http://stackoverflow.com/a/12141511/811740  .

boltons/mathutils.py
blebo: Fair comment. I've refactored the function names with the suffix `_from_iter` in https://github.com/mahmoud/boltons/commit/3263f4414d08a1a9ea406891112da3fbeb903196. 

I thought about including this code in the `iterutils` module, but I didn't for the following reasons:
1. These functions are alternatives to `math.ceil()` and `math.floor()` in the standard library.
2. These functions return a single value, not an iterable.
3. There may be others with similar utils that do not suit the `boltons` sub-modules available to date.

boltons/mathutils.py
blebo: Interesting. I have a look at this, and I'll try to come up with something more optimal.

blebo:
Apologies, I should have submitted this from a feature branch to begin with, as rebase has polluted the log above (I'm still getting my head around git). I think I'll resubmit the intended changes as a pull request, once I get re-organised.

PR #31, opened 2015-04-19
Properly parse tracebacks with missing source.  Resolves #30

### LINKED ISSUE #30: ParsedException.from_string produces wrong result when a frame is missing source (such as from eval)
Take this (modified from the original) trace:

```
  File "<string>", line 2, in _some_function
  File "myfile.py", line 3, in some_other_function
    return foo(bar, baz)
MyException: ExceptionValue
```

ParsedException gives me an incorrect result:

```
>>> ParsedException.from_string(x)
ParsedException(u'    return foo(bar, baz)', u'', frames=[{'source_line': u'File "myfile.py", line 3, in some_other_function', 'filepath': u'<string>', 'lineno': u'2', 'funcname': u'_some_function'}])
```

asottile:


asottile:
I guess I'll fix tbutils to work under python3?  Kinda unfortunate for the build to fail because I added tests :( (when it wouldn't have worked before then!)

mahmoud:
This looks like a rock solid PR. Big thanks for going the extra mile! Also, I've never seen a comment inside an if condition like that, but I'm too psyched about the fix to mind :100:

asottile:
<3

PR #32, opened 2015-04-20
Fix multiline exception messages

asottile:
Two fixes, sorry I bundled them into one PR, hope this is chill <3

### DIFF
--- a/boltons/tbutils.py
+++ b/boltons/tbutils.py
@@ -753,7 +753,7 @@ class ParsedException(object):
             frames.append(frame_dict)
 
         try:
-            exc_line = tb_lines[line_no]
+            exc_line = '\n'.join(tb_lines[line_no:])
             exc_type, _, exc_msg = exc_line.partition(':')
         except Exception:
             exc_type, exc_msg = '', ''
--- a/boltons/tbutils.py
+++ b/boltons/tbutils.py
@@ -754,7 +754,7 @@ class ParsedException(object):
 
         try:
             exc_line = '\n'.join(tb_lines[line_no:])
-            exc_type, _, exc_msg = exc_line.partition(':')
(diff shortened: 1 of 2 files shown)


mahmoud:
Looks good! No sweat at all. If the commit messages are descriptive, tests are there, and builds are passing, all bugfixing pull requests are welcome. :smile:

mahmoud:
So here's a question now. How common are multiline exceptions? For example, if we wanted to define a utility function for parsing through, say, a log file, looking for parseable Exceptions to parse, how would you suggest we do it in a multi-line exception-filled world?

Many (most?) exceptions this utility was meant to help parse will be embedded in other text with other lines. Some of which may be log lines or other exceptions, and I'm guessing very little of which will be additional lines of exception message. 

Unless we come up with a good solution for the practical context above, I may have to rethink merging that feature. A kwarg might be in order, perhaps.

asottile:
They're pretty common at Yelp where we have particularly descriptive exceptions such as: "Here's your error\n\nYou should do this: blah blah blah http://link/to/a/wiki/page".  We also have a lot of rethrown exceptions where we'll embed the original stacktrace in the new message (an unfortunate side-effect of being stuck on python2!)

I feel like if someone passes something that's not a traceback to this it should not guarantee any behaviour.  It already complains if the start of the text is non-traceback for instance :)

mahmoud:
Right. At one point already I've used this to jury rig a search through a log file that just tried to parse on every line and it worked out ok :P 

So in those Yelp-like cases (which I've seen a couple times externally, too), any thoughts on tokenizing between "traceback" and "non-traceback" text?

asottile:
Hard to say, I fortunately have well-formed data (a json blob with the formatted exception as a value)

PR #33, opened 2015-04-22
added sockutils with BufferedSocket and NetstringSocket

doublereedkurt:
in the interests of not letting the perfect be the enemy of the good, I'm going to go ahead and put out this pull request even though docs are still incomplete, and test coverage is not quite 100%

### DIFF
--- a/boltons/sockutils.py
+++ b/boltons/sockutils.py
@@ -36,7 +36,7 @@ class BufferedSocket(object):
         self.rbuf = ""
         return data
 
-    def peek(self, n, timeout=None):
+    def peek(self, n, timeout=_UNSET):
         'peek n bytes from the socket, but keep them in the buffer'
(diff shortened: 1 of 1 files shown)


mahmoud:
Yes, buffered sockets are definitely something I could see in the standard library. Right now this isn't passing tests on Python 3, but I think I can fix that. I'll manually merge and fix it locally.

doublereedkurt:
Oh right it uses str.join in a bunch of places.... That isn't going to fly
in Python 3

mahmoud:
Haha, you can say that again. Lots of stuff needed tweaking, but it now runs under Python 2.6, 2.7, 3.4, and PyPy. I'll be doing a lot more review, commenting, and documentation before it ends up advertised in the main docs. ;)

doublereedkurt:
Very nice :-)

PR #34, opened 2015-04-24
TaxonomyTree

rochacbruno:
``` python
from boltons.dictutils import TaxonomyTree

taxonomy = TaxonomyTree()

taxonomy.linguagens.script.python
taxonomy.linguagens.script.perl
taxonomy.linguagens.funcionais.elixir
taxonomy.linguagens.funcionais.f_sharp
taxonomy.databases.nosql.mongo
taxonomy.databases.nosql.cassandra
taxonomy.databases.sql.mysql
taxonomy.databases.sql.postgres
taxonomy['databases']['hybrid']['orient']



print taxonomy.as_dict()
{ 
    "databases": {
        "hybrid": {
            "orient": {}
        }, 
        "nosql": {
            "cassandra": {}, 
            "mongo": {}
        }, 
        "sql": {
            "mysql": {}, 
            "postgres": {}
        }
    }, 
    "linguagens": {
        "funcionais": {
            "elixir": {}, 
            "f_sharp": {}
        }, 
        "script": {
            "perl": {}, 
            "python": {}
        }
    }
}
```

### DIFF
--- a/boltons/dictutils.py
+++ b/boltons/dictutils.py
@@ -34,7 +34,8 @@ thanks to `Mark Williams`_ for all his help.
 .. _Mark Williams: https://github.com/markrwilliams
 """
 
-from collections import KeysView, ValuesView, ItemsView
+import json
+from collections import KeysView, ValuesView, ItemsView, defaultdict
 
 try:
     from itertools import izip
@@ -51,7 +52,7 @@ except ImportError:
 PREV, NEXT, KEY, VALUE, SPREV, SNEXT = range(6)
 
 
-__all__ = ['MultiDict', 'OMD', 'OrderedMultiDict']
+__all__ = ['MultiDict', 'OMD', 'OrderedMultiDict', 'TaxonomyDict', 'TaxonomyTree']
 
 try:
     profile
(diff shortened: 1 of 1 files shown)


rochacbruno:
Nice! now I know it is failing in Python 3x and PyPy, trying to fix.

mahmoud:
Hey Bruno, thanks for the interest and submission! If I'm to understand correctly then this is basically a defaultdict that creates instances of itself based on attribute accesses. Something like a tree and a [bunch](https://pypi.python.org/pypi/bunch/1.0.1) (and other libraries like it).

I guess my first question is, how this is specific to taxonomies?

blakev:
@mahmoud it makes sense given its other name of Phylogenic Chart, which is used to classify species. It looks [like this](http://bioweb.uwlax.edu/bio203/s2012/stauder_abig/website%20picture/TREE.jpg).

If this is still wanted I'm willing to fix it.

mahmoud:
Thanks for the generous offer @blakev, but I think at this point I'm struggling to see this as boltons functionality. As time has passed, the direction became clearer. To be a bolton, the code should arguably have been a builtin. And I'm not sure that TaxonomyTree meets that criteria. 

I agree it's a cool use of Python and I can probably come up with some uses, but I'd need to see some citations of where it is currently used (or would have been had boltons had it). As for phylogenic charts, I doubt people are constructing them through Python instead of loading a JSON/YAML file (or likely something more arcane).

PR #36, opened 2015-04-27
Mathutils (2nd attempt)

blebo:
New module for math utilities (supersedes Pull Request #29). Includes alternative ceiling/floor functions (incorporating feedback from @mgaitan).

### DIFF
--- a/boltons/mathutils.py
+++ b/boltons/mathutils.py
@@ -12,11 +12,11 @@ def ceil_from_iter(src, x, allow_equal=True):
         allow_equal (bool): Defaults to ``True``. Allows equality to be ignored if set to ``False``.
 
     >>> VALID_CABLE_CSA = [1.5, 2.5, 4, 6, 10, 25, 35, 50]
-    >>> ceil_from_iter(VALID_CABLE_CSA,3.5)
+    >>> ceil_from_iter(VALID_CABLE_CSA, 3.5)
     4
-    >>> ceil_from_iter(VALID_CABLE_CSA,4)
+    >>> ceil_from_iter(VALID_CABLE_CSA, 4)
     4
-    >>> ceil_from_iter(VALID_CABLE_CSA,4,allow_equal=False)
+    >>> ceil_from_iter(VALID_CABLE_CSA, 4, allow_equal=False)
     6
     """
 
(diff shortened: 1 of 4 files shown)


mahmoud:
Wow! Sorry, that was me noticing the very comprehensive testing and documentation.

I've definitely needed this functionality in the past. I wish `ceil` and `floor` took a second argument. Furthermore, `boltons` should probably have a `mathutils` going forward, as I'm sure others will have techniques to contribute, so this makes a very acceptable christening for that.

I'm a little iffy on the names used. For instance, can we just call the functions `ceil` and `floor`, but enhance the signature with an optional second argument; if left blank, the normal `ceil`/`floor` behavior is used. This would be a similar approach to `typeutils.issubclass()`, for instance. I'll have to sleep on it and ask around. If you have some thoughts do share. Until then, merged!

mgaitan:
Congrat @blebo

mahmoud:
So, after careful consideration and consultation of other devs, I've finalized the version of the API that will go public.

Rather than the long names, the functions are simply `ceil()` and `floor()` and they have the same API as the builtins, except they take an optional `options` argument.

The argument order was changed to make the target number first, in line with the existing `ceil` and `floor`. I think you'll agree that the `options` argument name is more semantic than the old `src` name. It also appropriately hints at the discretization effect it has on the mode of `ceil`/`floor`.

The `allow_equal` argument was eliminated because the built-in functions always allow equal. `ceil(4.0)` gives `4.0`. For the rare cases that require non-equal options only, this can be managed by removing the value from the options before calling the function.

The `sorted_src` argument was removed and sorting is always performed. Timsort is O(n) for sorted lists. I ran some tests and there would need to be many calls with many options to have a noticeable effect. While `sorted_src` felt like something I might use on the right day, it does not feel like something the standard library would have included. The more streamlined API makes these new functions feel like more powerful builtins instead of all-new functionality.

I've updated the architecture section of the docs to emphasize this preference for keeping APIs in the tone of the stdlib: https://boltons.readthedocs.org/en/latest/architecture.html#design-of-a-bolton

Finally, I updated the tests to use py.test, which was very easy thanks to their clearly-written structure. Take a look, I think you'll agree that pytest makes for more readable tests overall!

blebo:
Thanks for the merge, and the feedback. I agree, my orignal implementation was a bit clunky, and perhaps tried to address too many contrived corner cases. Tests look a lot cleaner - I'll definitely have to look into py.test further.

PR #38, opened 2015-05-01
`LRU.popitem` should return a key value pair

suzaku:


mahmoud:
Oh, very nice, thanks!

PR #39, opened 2015-05-18
create shortcuts for pairwise chunks since pairs (key/val) are common

ericfrederich:
Pairwise chunks are pretty common.


mahmoud:
Agreed that they are common, and these are the right names, but would you mind updating these to not be lambdas? It really helps with reprs and documentation/examples.

ericfrederich:
@mahmoud I changed them into functions and added them to the documentation.

mahmoud:
Great! Thanks!

PR #40, opened 2015-05-21
new is_container function

ericfrederich:
Hey, this is pretty useful.  I never use is_scalar but the effective inverse is_collection.

### DIFF
--- a/boltons/iterutils.py
+++ b/boltons/iterutils.py
@@ -58,6 +58,11 @@ def is_scalar(obj):
     """
     return not is_iterable(obj) or isinstance(obj, basestring)
 
+def is_container(obj):
+    """The inverse of :func:`is_scalar`.  Returns ``True`` if an object
+    is iterable but not a string type.
+    """
+    return is_iterable(obj) and not isinstance(obj, basestring)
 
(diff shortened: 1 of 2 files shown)


ericfrederich:
Any chance of getting this merged?

mahmoud:
You know, I was waiting for a real application to come along use it myself and feel it out, as I like to do with most contributed boltons, but this is simple enough, so sure why not. :+1:

ericfrederich:
There wasn't any real discussion about this.  I was wondering if is_container is a good name or if it should be is_collection.  I'd be okay with either.

mahmoud:
Hmm, is_collection is pretty good. I'll make that switch.

ericfrederich:
Thanks.
I use `is_collection` quite often when recursing over data structures which aren't strictly defined or of variable depth.  I was doing it today in fact and came to peek of this had been merged.  `not is_scalar` would work too but I think having both `is_scalar` and `is_collection` are useful since they both have meaningful names.

Please don't accept any merge requests for `uncallable` until something that isn't callable has a better name ;-)

mahmoud:
That sounds like sound advice to me :)

PR #42, opened 2015-06-17
Groupby iterator in the iterutils.py code

fivetentaylor:
I've found use for this function a lot!  I think it'd make a nice addition to the boltons library.

Thanks!

### DIFF
--- a/boltons/iterutils.py
+++ b/boltons/iterutils.py
@@ -398,3 +398,49 @@ def one(src, cmp=None):
                 return False
             the_one = i
     return the_one
+
+def groupby_iter(src, key=lambda x: x, keep=0):
+    """``groupby_iter()`` yields lists of neighboring elements
+    with equal keys from the iterable *src*
+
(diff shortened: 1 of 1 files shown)


mahmoud:
Hey Taylor, thanks for the submission! I see you've got a pretty healthy docstring on there already (reads pretty well), but given that there is a similar standard library function, how would you describe the differences between your groupby and [itertools.groupby](https://docs.python.org/2/library/itertools.html#itertools.groupby)?

fivetentaylor:
Hey Mahmoud!  I'm so sorry, there is virtually no difference other than the "keep" directive, which could just be accomplished by composition with another iterator.  I don't know how I've been missing that function all this time.  At first I thought maybe it was a new addition to itertools, but nope it's been there forever.  
Please disregard my pull request, hopefully I'll have better luck next time :)

mahmoud:
Hey, happens to the best of us, the standard library is huge! Looking forward to next time :)

PR #43, opened 2015-06-19
join function for iterutils

fivetentaylor:
So I'm trying again with my other favorite iterator function.  I've googled this time (somewhat) to make sure there's not an existing one out there (or right in itertools!)

The docstring describes it pretty well, let me know what you think!

Thanks,
Taylor

### DIFF
--- a/boltons/iterutils.py
+++ b/boltons/iterutils.py
@@ -399,56 +399,53 @@ def one(src, cmp=None):
             the_one = i
     return the_one
 
-def groupby(src, key=lambda x: x, keep=0):
+def join(iters, key=lambda x: x):
     """
     Description
     ===========
-    Iteration utility to group adjacent elements from an
-    iterator that share a common key
+    Iteration utility to join two or more iterators
+    sorted on a common key, very much like sql outer
+    join 
 
     Parameters
     ==========
-    src: an iterator or iterable object
-    key: a function that returns the key to group by from each
-    element in the iterator
-    keep: which elements to keep
-      - 0: all elements
-      - n < 0: last n elements
-      - n > 0: first n elements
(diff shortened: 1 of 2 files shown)


mahmoud:
FYI I've been thinking on and off about this for a while. You're definitely not the only one who've needed this before. A couple people I polled (and I) have even used in-memory sqlite tables for this basic functionality. So I'm :+1:, though I'm considering the implementation, API, and name. Just thought I'd let you know it's still on my mind. Will try to have a solid plan in the next week or so.

fivetentaylor:
Very cool!  I look forward to your decision, and definitely open to improvements on api/implementation.  I'll think on it some more as well.

doublereedkurt:
This is very interesting.  One thing I'd be curious about: have you encountered any practical use cases?  I can see how it might be a useful building block, but it is hard to think up a use case from scratch.

This seems pretty similar to the "zip" function.  Maybe it could be called zipby() or zipon() instead of join()?

I think join() might be easy to confuse with a concurrency primitive (e.g. Thread.join()  https://docs.python.org/2/library/threading.html#threading.Thread.join).

mahmoud:
Certainly thinking about it as a zip "on" a certain key function helps me think about it. If I had to guess, I think that'd be true of most Pythonists. I'd still definitely include the outer join comparison in documentation in the final product.

fivetentaylor:
I'm definitely open to changing the name.  In pandas they have concat and merge as well as join.  I think merge might be a reasonable choice.

doublereedkurt:
Pandas is an interesting comparison.  Everyone says good things about that
library.
I am becoming more and more convinced this will be a  handy function :-)

PR #44, opened 2015-07-30
Add first function

hynek:
Transplant from https://github.com/hynek/first , hereby donated. :)


mahmoud:
And graciously accepted! Thanks for the quick PR. 42 minutes elapsed since [initial tweet](https://twitter.com/hynek/status/626673067884998656). I'll cut a new version before bed.

hynek:
FWIW you said something about crediting myself but I haven’t found the right place. ¯\_(ツ)_/¯

mahmoud:
Ah yes, you're probably looking for a CONTRIBUTORS or something. I've just been doing it in the docstrings. Closer to the code means it'll get pasted around more. ;)

PR #46, opened 2015-08-11
Handle teens in strutils.ordinalize

zeroSteiner:
This PR changes `strutils.ordinalize` to correctly handle numbers in the 1x decade. Prior to this numbers such as 12, and 111 were being incorrectly given the nd and st extensions respectively. This checks if the character before the last is a 1 and uses th as the extension.


mahmoud:
Great catch! I may tweak a couple lines of the implementation, but this looks good overall. Thank you!

PR #47, opened 2015-09-18
Missing docs for unique and unique_iter

kevinastone:
Went to go write the missing `unique` iterutil and saw it was already there, just undocumented.


mahmoud:
Oh, good catch, thank you! Hope that wasn't too much of an inconvenience, thanks again!

PR #49, opened 2015-10-06
iterutils.backoff: Added support for jitter.

suranap:
Just so people don't have to reinvent the wheel, I added simple support for jitter in the backoff and backoff_iter functions. I also added a simple version of full_jitter, based on this [blog post from AWS](http://www.awsarchitectureblog.com/2015/03/backoff.html).

### DIFF
--- a/boltons/iterutils.py
+++ b/boltons/iterutils.py
@@ -16,6 +16,7 @@ __all__ = ['is_iterable', 'is_scalar', 'split', 'split_iter', 'chunked',
 
 import math
 import itertools
+import random
 from collections import Mapping, Sequence, Set, ItemsView
 
 try:
@@ -336,7 +337,7 @@ def frange(stop, start=None, step=1.0):
     return ret
 
 
-def backoff(start, stop, count=None, factor=2):
+def backoff(start, stop, count=None, factor=2, jitter=lambda x:x):
     """Returns a list of geometrically-increasing floating-point numbers,
     suitable for usage with `exponential backoff`_. Exactly like
     :func:`backoff_iter`, but without the ``'repeat'`` option for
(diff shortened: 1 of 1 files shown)


mahmoud:
Hi Pinku! Thanks for the contribution. I assume you saw the note about jitter in the docstring?

I've been considering your API. A simple callback is straightforward, but the approach I've been thinking about is a more composable generator that would perform randomization on each of the elements as yielded from an iterator. That way it could be used with numeric sequences not yielded from the backoff generator. What do you think?

suranap:
Hi,

Technically it's all the same and we're just discussing API design. As long as you're consistent it doesn't matter. I'll offer my 2 cents. You have 3 obvious choices: 

1) Embed jitter in backoff as I did:
```python
def backoff(start, stop, count=None, factor=2, jitter=lambda x:x):
```
2) Provide a function as you recommend:
```python
def add_jitter(iter):
    return map(full_jitter, iter)

x = add_jitter(backoff(1,10)) 
```
3) Let people just do it themselves:
```python
[full_jitter(i) for i in backoff(1,10)]
```

IMHO, jitter is so important to backoff that it should be the way people use it. Therefore, (1) makes it glaringly obvious to users. I should have made `full_jitter` the default and force users to turn it off with `None`. If someone wants to jitter a different int stream they can easily do (3). So my hunch is that (2) would be of limited value. 

I'm cool with whatever you decide. Also, you might want to add 1 or 2 more jitter functions. `full_jitter` looks lonely as the only option.

mahmoud:
Hi again @suranap, it's been a long time, but I have put a lot of thought into it, and I'm happy to say [iterutils.backoff now has jitter](https://boltons.readthedocs.org/en/latest/iterutils.html#boltons.iterutils.backoff_iter). 

After much usage and discussion, I've opted to keep the API simple. Jitter is disabled by default. The functionality you desired is achievable by passing `jitter=True`. The "advanced" use case is to pass a float -1.0 < x < 1.0, to treat the curve as a min or a max. 

I plan on preparing a separate blog post about conceptualizing jitter in production, and why there is no such thing as a generic jitter function. Thanks again for contributing, and emphasizing the importance of this feature!

PR #51, opened 2015-11-05
Fixdoc

mgaitan:
there was a couple of malformed links in `statutils.py`'s docstrings. 

Also, `funcutils` seems to have missing tests, because tox passed but a `get_module_callables()` misses `basestring` in py3


mahmoud:
Looks great, thank you!

PR #56, opened 2016-01-25
Small format fix in iterutils.one for None

nicoddemus:


mahmoud:
Thanks for helping make the docs (and thus all of boltons) better! :)

nicoddemus:
Sure thing, thanks for the excellent library! :smile:

PR #57, opened 2016-02-16
Add is_uuid util.

kdeldycke:
Not sure if this tiny utility needs its own `uuidutils.py` submodule so I added it to `strutils.py` by default, as it works on both strings and UUID objects.

### DIFF
--- a/boltons/strutils.py
+++ b/boltons/strutils.py
@@ -8,6 +8,7 @@ provided by ``strutils``.
 from __future__ import print_function
 
 import re
+import uuid
 import zlib
 import string
 import unicodedata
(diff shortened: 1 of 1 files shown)


mahmoud:
Hey there! Thanks for the contribution. Looks pretty good at first blush. :) I'll dig into this a bit more shortly. Thanks again!

mahmoud:
(and yes, strutils is the right place)

kdeldycke:
Thanks again for the fast merge ! :)

mahmoud:
Yep no problem. I'll cut a release shortly.

kdeldycke:
That's even better ! :smiley_cat:

PR #58, opened 2016-02-16
Add is_ascii util.

kdeldycke:
In the same spirit as #57, here is a tiny ``is_ascii()`` util.

### DIFF
--- a/boltons/strutils.py
+++ b/boltons/strutils.py
@@ -414,6 +414,23 @@ def asciify(text, ignore=False):
         return ret
 
 
+def is_ascii(text, ignore=False):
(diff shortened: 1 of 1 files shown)


mahmoud:
I see some tests aren't passing, but no worries. I think there is a faster and more standalone way to to do this other than calling asciify. I'll do some tests tonight. Good idea, nonetheless. Thanks!

mahmoud:
All merged up! If you'd like, you can verify my updates. Thank you again!

kdeldycke:
Your adjustments looks good. Thanks for the fast merge ! :)

PR #59, opened 2016-02-17
Add an util to the fractional precision of a number.

kdeldycke:


mahmoud:
Hmm, can you explain how this is intended to be used? I ask because '0.110' to me seems to have a precision of 3, but precision() returns 2. That seems to be leaving valuable significant digits on the table.

kdeldycke:
If I was aiming for the result you suggested, then a kind of `len(str(number).split('.')[1])` would be enough.

But that's not what I was looking for. Let me try to explain the implementation around the `Decimal`.

For context, I coded this utility for an accounting/billing pipeline, in which precision is a non-negotiable requirement.

Arithmetics with `Decimal` may ends up with a number of arbitrary digits being tracked, even if said arithmetics are designed to converge to the expected, lower precision. See for example:
```python
>>> from decimal import Decimal
>>> Decimal('100.11') * Decimal('10')
Decimal('1001.10')
```

In which case the resulting `Decimal('1001.10')` is visually of precision 2 but mathematically of precision 1, because of the trailing `0`.

My `precision()` method is used here to get the normalized lower precision of a number. Which, believe me, helps a lot when it's time to deal with money and want to ensure no micro-dollars slips through the cracks of floating point computation.

Not sure if I'm making the utility of this method really clear. But in case of doubt, we can rename it `minimal_precision` or similar.

kdeldycke:
Aha! Also noticed that I used this `precision()` method when I proposed a monkey-patch to deal with `babel`'s loss of precision: https://github.com/python-babel/babel/issues/90#issuecomment-174456874

But that's not the real use-case of the method, just a side-note.

mahmoud:
Hehe, believe me, I want to understand. It's just, I've worked at PayPal for many years, with a lot of numbers, including the money-representing kind, and we use Decimal with no issues. There is an explicit rounding at times, but "normalized lower precision" doesn't ring any bells. 

In short, to my knowledge, we cannot infer the number of significant digits from any one representation of a number. It has to be separately tracked and explicitly rounded. Sorry for all the questions, I'm just hesitant to merge something that might mislead the less numerically inclined among us. :)

mahmoud:
To illustrate, why is `precision(0.5 + 0.6)` higher than `precision(0.5 + 0.5)`?

kdeldycke:
Maybe it's in the semantics then. My method might not implement the generally accepted definition of a precision. In which case it truly is a method that is "normaliz[ing] lower precision" as you put it.

If you want to play it safe, what do you think of rewriting the precision utility according your specs, then try to add my additional normalizing behaviour as an option? Something like:
```python
def precision(number, normalize=False):
```

Another direction might be to add a kind of truncation companion utility.

mahmoud:
I've thought about it quite a bit, and shown it to a few folks. 

Unfortunately it seems too unintuitive, both to me, and others. The other PRs I merged because there had been times I could have used them. This one, it's too much guessing. 

I may reconsider, if this ever comes up, but really I think this should only work with Decimal objects, in which case this becomes clearer as a one liner using the Decimal directly.

Still, don't let this discourage you from bringing more utils! :)

kdeldycke:
Ok, as you wish. And thanks for the fast and detailed feedback. I really appreciate the discussion, as I find this behaviour quite rare these days. Thanks for the friendly and extensive maintainership! :)

Oh, I'll still try to offload my internal utils to boltons: your project is a fantastic initiative and a time-saver! :)

kdeldycke:
For those wondering, the one-liner @mahmoud is referring to ends up being:
```python
>>> from decimal import Decimal
>>> abs(min(0, Decimal('12133.123000').normalize().as_tuple().exponent))
3
```

PR #61, opened 2016-02-23
On Windows, try to remove the path before renaming.

### LINKED ISSUE #60: os.rename called in fileutils breaks on Windows if directory exists
This was seen in the pip-tools code that is vendoring the AtomicSaver code from boltons.

https://github.com/nvie/pip-tools/pull/322

in essence, the os.rename cannot be implemented atomically on Windows, but the current code breaks if exercised. (see the issue in pip-tools). 

I will make a PR here as suggested by https://github.com/nvie

davidovich:
This cannot be made atomic on Windows, but it is better than raising an exception on usage.

Fixes #60

### DIFF
--- a/boltons/fileutils.py
+++ b/boltons/fileutils.py
@@ -225,9 +225,20 @@ def atomic_save(dest_path, **kwargs):
     return AtomicSaver(dest_path, **kwargs)
 
 
+# rename operation fails on Windows when destination exists
+# see http://stackoverflow.com/questions/8107352/force-overwrite-in-os-rename
+if os.name == 'nt':
+    def _rename(old_path, new_path):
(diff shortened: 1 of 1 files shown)


mahmoud:
Thank you for the bug report and the initial implementation. I think the current implementation linked from #61 ([link](https://github.com/mahmoud/boltons/blob/master/boltons/fileutils.py#L237)) has some desirable advantages in terms of atomicity. I've tested it on my Windows machine and would be very obliged if you did the same! :)

davidovich:
@mahmoud Excellent, thank you, this better implementation fixes the problem. I will report to pip-tools the good news.

Fixed and verified.

PR #63, opened 2016-02-29
Improved the build matrix

thedrow:
Added caching and python 3.5 to the build matrix.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,11 @@
 language: python
+sudo: false
+cache: pip
 python:
   - "2.6"
   - "2.7"
   - "3.4"
+  - "3.5"
   - "pypy"
 
 install: "pip install -r requirements-test.txt"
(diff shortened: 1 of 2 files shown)


mahmoud:
Hey! Thanks for the PR! 

The results are interesting. As @hynek mentioned, 3.5 appears superborked! A new version of py.test might add the proper ast handling?

Is there any reason to cache? Boltons has few-to-no deps and build times were never the issue. An unnecessary cache is often the cause of a future bug.

The `sudo: false` is also interesting, what's the rationale there? 

Thanks again!

hynek:
`sudo: false` gives you the new (faster) Docker-based infra.  Works only if you don’t need root privileges though.

thedrow:
I can remove the cache if you don't want it. It does save you the trouble of downloading tox & pytest.

mahmoud:
@hynek: Good to know!

@thedrow Allll green, nice. I'll give the cached version a whirl and see how it feels. If the productivity is worth it, it might stick around. Thanks again!

thedrow:
No problem.

PR #64, opened 2016-03-02
Fix is_uuid() on un-parsable types.

kdeldycke:
`boltons.strutils.is_uuid()` method raise an exception with garbage types. I expect here `is_uuid()` to return `False` for un-parsable types.

The first commit provide evidence of the unwanted behaviour.
The second commit provide a fix.

### DIFF
--- a/boltons/strutils.py
+++ b/boltons/strutils.py
@@ -700,7 +700,7 @@ def is_uuid(obj, version=4):
     if not isinstance(obj, uuid.UUID):
         try:
             obj = uuid.UUID(obj)
-        except (TypeError, ValueError):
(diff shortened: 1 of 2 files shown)


mahmoud:
Makes sense, thanks for looking after this!

kdeldycke:
Unit-tests on my proprietary code base caught it. Thanks for the really fast merge! :smiley:

Any chance for a bug-fix release soon?

mahmoud:
(the bug fix was pushed over the weekend, new versions of boltons are good to go!)

kdeldycke:
Thanks for releasing boltons 16.1.1 ! :)

PR #65, opened 2016-03-02
Update trove classifers.

kdeldycke:
Just a little cosmetic change.


mahmoud:
Ah yes, easy to forget, but good to keep up to date all the same. Thanks!

PR #68, opened 2016-03-24
Remove unused test files

wzyboy:
Looks like these two files are not referenced other where in the project.


mahmoud:
Artifacts of an earlier time. :) Thanks!

Issue #66, opened 2016-03-08
Cached property


kdeldycke:
I'm currently using [`cached-property`](https://pypi.python.org/pypi/cached-property) alongside `boltons` and I feel like `cached-property` features can be emulated with `boltons.cacheutils`.

Is this under the scope of `boltons`? Should we add some utils to support this feature in `boltons`?

mahmoud:
I agree that that sort of functionality is a decent fit for cacheutils, and I have done a few specific implementations of the pattern so looking at a bolton makes sense. That said, I don't think there's anything in particular already in cacheutils that would enable it, so it's not the lowest-hanging fruit.

Is there something specific that you'd like to see in a new cached property API/implementation?

Thanks for bringing it up though, this is definitely something that comes up regularly!

kdeldycke:
Thanks for the feedback. The way I cache properties is dead simple, stupid. So I have no particular requirement.

As you said it's quite a common pattern. Just implement a minimal working cached properties based on your needs. We'll iterate from there depending on real use-cases.

mahmoud:
Hey @kdeldycke thanks for your patience. Have a look at [the new cachedproperty](https://github.com/mahmoud/boltons/blob/c74252986b1fbf89d651f1156127c4077d3aa3ef/boltons/cacheutils.py#L621). It takes the standard approach I've used in a couple of projects. There may be some niceties worth adding, but this one is the one that works for me!

kdeldycke:
The new `cachedproperty()` is nice! That's exactly what I was looking for. I'll try it in the next couple of days. Let's close the issue for now. I'll re-open this one or a new one if I stumble upon any issue.

Thanks @mahmoud ! :)

mahmoud:
Great! Planning to do a release this week.

PR #69, opened 2016-05-06
Make cache accept callable arguments, add tests.

grundic:
Hi, I've added possibility to use cache as callable and added some tests.

### DIFF
--- a/boltons/cacheutils.py
+++ b/boltons/cacheutils.py
@@ -450,7 +450,7 @@ def _make_cache_key(args, kwargs, typed=False, kwarg_mark=_KWARG_MARK,
 class CachedFunction(object):
     def __init__(self, func, cache, typed=False):
         self.func = func
-        self.cache = cache() if hasattr(cache, '__call__') else cache
+        self.cache = cache() if callable(cache) else cache
         self.typed = typed
 
     def __call__(self, *args, **kwargs):
--- a/boltons/cacheutils.py
+++ b/boltons/cacheutils.py
@@ -450,7 +450,7 @@ def _make_cache_key(args, kwargs, typed=False, kwarg_mark=_KWARG_MARK,
 class CachedFunction(object):
     def __init__(self, func, cache, typed=False):
         self.func = func
-        self.cache = cache
+        self.cache = cache() if hasattr(cache, '__call__') else cache
         self.typed = typed
 
     def __call__(self, *args, **kwargs):
--- a/tests/test_cacheutils.py
+++ b/tests/test_cacheutils.py
@@ -2,7 +2,9 @@
 
 import string
 
-from boltons.cacheutils import LRU, LRI
+import mock
+
+from boltons.cacheutils import LRU, LRI, cached
 
 
 def test_lru_add():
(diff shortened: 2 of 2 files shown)


boltons/cacheutils.py
```diff
@@ -450,7 +450,7 @@ def _make_cache_key(args, kwargs, typed=False, kwarg_mark=_KWARG_MARK,
 class CachedFunction(object):
     def __init__(self, func, cache, typed=False):
         self.func = func
-        self.cache = cache
+        self.cache = cache() if hasattr(cache, '__call__') else cache
```
mahmoud: this should almost certainly be using [callable](https://docs.python.org/2/library/functions.html#callable)

tests/test_cacheutils.py
```diff
@@ -2,7 +2,9 @@
 
 import string
 
-from boltons.cacheutils import LRU, LRI
+import mock
```
mahmoud: Tests look pretty good, but the usage of mock seems excessive. We don't need mock functions, a normal function returning a constant should do just fine.

kdeldycke:
Is this PR covers #66 too?

mahmoud:
@kdeldycke no, not at all :stuck_out_tongue: soon though!

@grundic upon further review, don't you think this would make more sense in the `__call__`?

grundic:
@mahmoud, yep definitely checking for `callable` should be done via your proposal, fixed that.
What about `mock` - sure I can create simple nested functions and test them, but how do I check call count? It could look something like [this](http://stackoverflow.com/questions/21716940/is-there-a-way-to-track-the-number-of-times-a-function-is-called), but it would make tests a little bit complicated how do you think?

mahmoud:
@grundic My more-important point is that I'm not sure it makes sense to call the cache argument in the `__init__`. If you do that, the user might as well just call it before they pass it in. Instead, the callable should be called in the `cached` decorator's `__call__`.

Still not sure if that sort of lazy access is a good pattern, but it's probably not harmful. I would guess there are precious few callable cache instances that don't return caches.

grundic:
If cache as a function would be in `__call__` it would executed each time the cache is requested. Or there should be some check whether we already executed and getted the cache. In `__init__` we just initializing the value one time.

I don't think that users could easily call the function and pass the value to the decorator, as it evaluated at class/method creation. So it would not be possible to dynamically configure cache at runtime. With function as argument it is possible. Actually it was my naive approach to implement `cachedmethod` pattern with access to cache with `operator`:
```python
from boltons.cacheutils import LRU,cached
import operator

class ClassWithCache(object):
    def __init__(self):
        self.cache = LRU()

    @cached(operator.attrgetter('cache'))
    def number(self, value):
        return value

instance = ClassWithCache()
print(instance.number(19))
```

But as I discovered, it requires a lot more implementation, so I created #70 :)

mahmoud:
I figured it would be easier to show than tell, so I went ahead and implemented this. [Have a look](https://github.com/mahmoud/boltons/blob/c74252986b1fbf89d651f1156127c4077d3aa3ef/boltons/cacheutils.py#L464), maybe take it for a spin, let me know what you think!

grundic:
Excellent, closing my PR.
Thank you!

PR #74, opened 2016-05-10
fix #72 fileutils doctest error on Windows

### LINKED ISSUE #72: fileutils error on Windows
A follow-on from https://github.com/mahmoud/boltons/issues/71

python 2.7, 3.4, and 3.5--both 32-bit and 64-bit--have a test failure in the fileutils.

    _________________ [doctest] boltons.fileutils.iter_find_files _________________
    462             search. Yielded paths will include this as a prefix. 
    463         patterns (str or list): A single pattern or list of 
    464             glob-formatted patterns to find under *directory*. 
    465         ignored (str or list): A single pattern or list of 
    466             glob-formatted patterns to ignore.
    467  
    468     For example, finding Python files in the current directory: 
    469  
    470     >>> filenames = sorted(iter_find_files(_CUR_DIR, '*.py')) 
    471     >>> filenames[-1].split('/')[-1] 
    Expected:
        'typeutils.py'
    Got:
        'C:\\Miniconda-x64\\conda-bld\\work\\boltons\\typeutils.py' 
     
    C:\Miniconda-x64\conda-bld\work\boltons\fileutils.py:471: DocTestFailure 

Looks like a simple fix to use `os.path.basename` there instead of `.split('/')`.

I'll work on a PR at some point if somebody else doesn't beat me to it.

kalefranz:


mahmoud:
Good catch! Though part of me wonders why I didn't go with that from the beginning. Thanks!

PR #75, opened 2016-05-17
Add functions strpdate() and daterange(), which make working with date objects more convenient.

kiwidrew:
I've often wanted to use a `strptime()`-style format string to parse date values, and find that having a complimentary `strpdate()` function is clearer than writing e.g. ```datetime.strptime(val, '%Y-%m-%d').date()``` all over the place.

Also, iterating over date ranges can be quite awkward, because any date arithmetic requires the involvement of the dreaded `timedelta` class.

### DIFF
--- a/boltons/timeutils.py
+++ b/boltons/timeutils.py
@@ -23,7 +23,7 @@ degree of accuracy in corner cases, check out `pytz`_.
 import re
 import time
 import bisect
-from datetime import tzinfo, timedelta, datetime
+from datetime import tzinfo, timedelta, date, datetime
 
 
 def total_seconds(td):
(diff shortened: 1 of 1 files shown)


mahmoud:
These are good ones! Nice. A bit of feedback:

* I think the extra value checks in `strpdate` are kind of extraneous. If someone wants to parse a date out from a string with hhmmss in it, let them, I say. Of course removing that turns it into a trivial function. But still one I feel like would have made for improved parallelism for the `datetime.date` type. So still good and mergeable
* `daterange` is also good. That said, I see some issues. 
    * The first is that `assert`s are kind of bad form in this case. They can be disabled and don't provide great error messages. Explicit checks with good `ValueError`s and `TypeError`s are preferable.
    * The second is that day-only ranges are tough for iterating through months and years. Leap years throw it off. And months are of course all different lengths. So we could either complicate this function a lot, or keep it simple. A name change. How about, `dayrange`?

Thanks for the great work!

kiwidrew:
Thanks for the excellent (and quick) feedback!  I really like to idea behind your `boltons` module, and these are two functions that I have been carrying around with me across various projects.  I'll make some changes based on your suggestions and then we can take another look.

mahmoud:
Great! Yes, in fact, ideally, I'd like to include these in the next release, so if it's not likely you'll be able to make the changes in the next day or so, I can merge and modify them myself. It's really no trouble.

kiwidrew:
I'm going to argue for keeping the name `daterange`, as this is a `range`-alike function that operates on `date` values.  In the future I would also like to contribute a `datetimerange` function which can likewise operate on `datetime` values.  I think month and year iterators are out-of-scope, because Python's `timedelta` object only supports intervals composed of `(day, seconds, microseconds)`.

kiwidrew:
I'll be unavailable for the next couple of days, so please feel free to make any additional changes that you think are required – I trust your judgement here.

mahmoud:
So, I made a few tweaks, here are the highlights:

* Year and month are now supported with a reasonably compact syntax
* Negative steps are also supported (including with year and month)
* The stop argument is now required (though None is still supported for infinite sequences, a bit too dangerous of an implicit behavior)

Have a look at the docs and take it for a spin! http://boltons.readthedocs.io/en/latest/timeutils.html#boltons.timeutils.daterange

PR #76, opened 2016-05-23
Added range2list() and list2range() functions to strutils.py

blebo:
These functions will be useful for working with custom page range strings that are typically used in printer dialog boxes.

Tested on OSX 10.11.5 with Python 3.5.1 (from Homebrew).

### DIFF
--- a/boltons/strutils.py
+++ b/boltons/strutils.py
@@ -31,7 +31,7 @@ __all__ = ['camel2under', 'under2camel', 'slugify', 'split_punct_ws',
            'asciify', 'is_ascii', 'is_uuid', 'html2text', 'strip_ansi',
            'bytes2human', 'find_hashtags', 'a10n', 'gunzip_bytes',
            'iter_splitlines', 'indent', 'escape_shell_args',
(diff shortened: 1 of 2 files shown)


mahmoud:
Hey! Nice! I like this functionality pretty well, and thanks for putting in the hybrid fixes. I'm a bit iffy on the name, though a better one doesn't really leap to mind. `range2ints`/`ints2range` maybe? Doesn't really roll off the tongue, I'll admit.

blebo:
Cheers. Yeah a name change works for me. My other suggestion would be `range_string2ints()`/ `ints2range_string()` - as it is a bit more descriptive and avoids the direct association with the built-in `range()` function compared to the original function names. What do you think?

mahmoud:
How about `parse_int_list` and `format_int_list`? It doesn't match the `x2y` naming scheme, but I'm not too attached to that anyways.

mahmoud:
Great, all merged up. I made a couple docstring changes, but nothing major. I'll be releasing a new boltons tonight or tomorrow, so look for it then. Thanks again!

PR #77, opened 2016-05-24
Function builder

markrwilliams:
Python 3 compatibility

### DIFF
--- a/tests/test_funcutils_fb.py
+++ b/tests/test_funcutils_fb.py
@@ -1,0 +1,35 @@
+
+import pytest
+
+from boltons.funcutils import wraps
+
+
+def test_wraps_basic():
+
+    def pita_wrap(flag=False):
+
+        def cedar_dec(func):
+            @wraps(func)
+            def cedar_wrapper(*a, **kw):
+                return (flag, func.func_name, func(*a, **kw))
+            return cedar_wrapper
+
+        return cedar_dec
+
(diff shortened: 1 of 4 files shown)


tests/test_funcutils_fb.py
```diff
@@ -0,0 +1,167 @@
...
+    else:
+        exec_ = __builtins__["exec"]
+        exec_(code, globals(), names)
+
+
+@pytest.mark.skipif(IS_PY2, reason="Contains Python 3-only syntax")
+def test_wraps_py3():
+    F = {}
+
+    def C(src):
```
mahmoud: Very nice!

mahmoud:
Very nice work. This is a pretty complex area and I think this is close to the best that can be done. I like that you managed to work in the non-py2 syntax in the same test file, though I had guessed that would have gone in a second file/directory and then some conditional pytest discovery stuff would happen. Do you think that would be better? (syntax highlighting, linting, etc.)

markrwilliams:
Glad you like it!

Good call with the code-in-strings nonsense - pytest provides
http://pytest.org/latest/writing_plugins.html#_pytest.hookspec.pytest_ignore_collect
which is close enough to what you described.

mahmoud:
Well, with the keyword-only marker brittleness handled and tested, plus Python 3 tests split out, I think that all is looking good! Merging as soon as CI passes.

PR #93, opened 2016-09-20
Feature: Return namedtuples for get_format_args

wcooley:
The nested tuple/list structures that are returned by `get_format_args` are fairly cumbersome; returning data `namedtuple` makes accessing the data much cleaner.

I am a bit uncertain about the field names themselves - - I chose 'positional' and 'named' because the documentation for the function describes them that way, but I chose 'key' and 'index' for the component (rather than 'position' and 'name') because those are the standard way to refer to them. It seems a bit inconsistent, however -- perhaps the `FormatArgs` fields should be `indexed` and `keyed`? (or "...s" instead of "...ed"?)

### DIFF
--- a/boltons/formatutils.py
+++ b/boltons/formatutils.py
@@ -41,6 +41,7 @@ from __future__ import print_function
 import re
 from string import Formatter
 
+from .namedutils import namedtuple
 
 __all__ = ['DeferredValue', 'get_format_args', 'tokenize_format_str',
            'construct_format_field_str', 'infer_positional_format_args',
@@ -117,6 +118,10 @@ _TYPE_MAP = dict([(x, int) for x in _INTCHARS] +
                  [(x, float) for x in _FLOATCHARS])
 _TYPE_MAP['s'] = str
 
+FormatArgs = namedtuple('FormatArgs', ('positional', 'named'))
+PositionalFormatArg = namedtuple('PositionalFormatArg', ('index', 'type'))
+NamedFormatArg = namedtuple('NamedFormatArg', ('key', 'type'))
+
 
 def get_format_args(fstr):
     """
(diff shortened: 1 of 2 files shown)


wcooley:
Also, I tested this locally with py27, py35 and pypy. (I don't have py34 and I couldn't get the py26 virtualenv to build.)

boltons/formatutils.py
```diff
@@ -41,6 +41,7 @@
 import re
 from string import Formatter
 
+from .namedutils import namedtuple
```
mahmoud: can you change this to:

```python
try:
    from boltons.namedutils import namedtuple
except ImportError:
    from collections import namedtuple
```

This will allow the module to continue to be used both inside and outside of the boltons package.

tests/test_formatutils.py
```diff
@@ -48,6 +52,40 @@ def test_get_fstr_args():
...
+        ([], [('hello', str), ('fchar', str), ('width', str)]),
+        ([(0, str), (1, int), (2, float)], []),
+        # example 6 is skipped
+]))
+def test_get_format_args(sample, expected):
+    """Test `get_format_args` result as tuples."""
+    assert get_format_args(sample) == expected
+
+
+@pytest.mark.parametrize(('sample', 'expected'), zip(_TEST_TMPLS, [
```
mahmoud: hehe this inline parametrize is kind of long, just curious, is this common practice with pytest?

boltons/formatutils.py
wcooley: Done. It also occurs to me that I have also neglected to update the documentation.

tests/test_formatutils.py
wcooley: I assume so; the pytest docs show an [example](http://docs.pytest.org/en/latest/parametrize.html) of a multi-line parametrize right at the top. There are also a few instances in the pytest tests themselves (such as in *test_conftest.py*, *test_config.py* and *test_mark.py*).

If I have something like this that I intend to reuse exactly as-is, rather than moving the list of samples to a separate variable, I create a new mark:
```
pytest.mark.get_format_samples = pytest.mark.parametrize(('sample', 'expected), ...
```

mahmoud: Interesting, thanks!

PR #94, opened 2016-09-21
Add coerce_to_json function

wcooley:
This is a function that tries to map data to one of the handful of primitive types supported by JSON. I have included tests using it both for the *default* parameter of `json.dump` and with `remap`.

There is a bit more work to be done but I wanted to first see if there way any interest in it before I put more effort into it.  (I am also open to suggestions for the name.)

### DIFF
--- a/boltons/jsonutils.py
+++ b/boltons/jsonutils.py
@@ -11,7 +11,15 @@ from __future__ import print_function
 
 import os
 import json
+from collections import Iterable, Mapping
+from itertools import product
 
+try:
+    from collections import UserDict
+except ImportError:
+    from UserDict import UserDict
+
+from boltons.iterutils import first, is_scalar
 
 DEFAULT_BLOCKSIZE = 4096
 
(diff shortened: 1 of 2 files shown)


mahmoud:
Absolutely! This is certainly very interesting! I'm curious, can you go into a bit of detail about where and how you've applied this technique?

As for structural changes, a minor point would mostly be to "vendor in" the `first` and `is_scalar` implementations from `iterutils` to comply with [the integration architecture](http://boltons.readthedocs.io/en/latest/architecture.html#integration).

wcooley:
I tend to use *set*s for containers and (especially for test fixtures) [types.MappingProxyType](https://docs.python.org/3.5/library/types.html#types.MappingProxyType), the read-only dict proxy, both of which play havoc with JSON. I am also lazy to the point of masochism and will spend hours developing a general-purpose solution rather than do the same thing twice.

For example, I have a client for an HTTP API that updates groups given a list of identifiers to either add or remove, so I build the list of identifiers that should be present (from a source which will have duplicates), request the current identifiers in the group, do set operations for the adds and removes and then throw the sets of adds and removes at my client. The only downside is that instead of passing the data to *requests.requests* as `requests.requests(..., json=data, ...)`, I have to use
```
json_data = json.dumps(data, default=coerce_to_json)
requests.post(..., data=json_data, headers=['content-type': 'application/json'], ...)
```
Alternatively, I could use `remap` to convert the data, and did at first, but I felt like I figured out how to make it work by stabbing at it until it bled, rather than actually understanding what was going on, which I did not think would be as maintainable.

wcooley:
For the vendoring-in of *first*, *is_scalar* (and consequently *is_iterable*), would you prefer them to be:
1. Defined in the module namespace or within the function itself?
1. Re-named with a leading underscore?
1. Imported and only used on *ImportError*, like *namedtuple* in #93?
1. Copied wholesale with docstring (which might might comparing for future updates easier), copied without the docstring but the function body as-is or minimized (e.g., for *is_scalar* I'd copy (*is_iterable* and switch the *return* values)?

So many questions for so few lines... Guess I should research what's already being done.

wcooley:
I looked at the existing functions and did not find too many examples other than *make_sentinel*, so I've done what seems best to me; let me know if this is the preferred approach.

PR #96, opened 2016-09-27
Simplified and optimized iterutils.first.

coady:


mahmoud:
Very nice! Talk about a reduction! Pretty sure I'm ok with this because that function has pretty decent test coverage, but might be curious what @hynek thinks. Either way, nice work, @coady!

hynek:
Checking back, the proposed version is in first’s README and I haven't used it *in* first because first was Python 2.5-compatible when it came out.

mahmoud:
I'll take that a +1, then! boltons.iterutils is a 2.6+ shop. I think 2.6 added next() which made this oneliner possible. :) Thank you both!

hynek:
I would like to encourage you to drop 2.6. ;)  (and 2.7! come to the bright side!)

mahmoud:
If we're talking brightness, it's hard to find a brighter sun than *compatibility*. :)

PR #97, opened 2016-09-30
Simplified and optimized iterutils.bucketize.

coady:
Old:

```
In []: timeit iterutils.bucketize(range(10 ** 6), (2).__rmod__)
1 loop, best of 3: 304 ms per loop
```

New:

```
In []: timeit iterutils.bucketize(range(10 ** 6), (2).__rmod__)
1 loop, best of 3: 216 ms per loop
```

And the removed `TypeError` checks will occur anyway.

### DIFF
--- a/boltons/iterutils.py
+++ b/boltons/iterutils.py
@@ -13,7 +13,7 @@ following are based on examples in itertools docs.
 import math
 import random
 import itertools
-from collections import Mapping, Sequence, Set, ItemsView
+from collections import Mapping, Sequence, Set, ItemsView, defaultdict
 
 try:
     from typeutils import make_sentinel
@@ -480,18 +480,12 @@ def bucketize(src, key=None):
     item. See :func:`partition` for a version specialized for binary
     use cases.
     """
-    if not is_iterable(src):
-        raise TypeError('expected an iterable')
     if key is None:
         key = bool
-    if not callable(key):
-        raise TypeError('expected callable key function')
-
-    ret = {}
+    ret = defaultdict(list)
     for val in src:
-        keyval = key(val)
-        ret.setdefault(keyval, []).append(val)
-    return ret
+        ret[key(val)].append(val)
+    return dict(ret)
 
 
 def partition(src, key=None):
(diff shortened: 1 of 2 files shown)


mahmoud:
Almost 30%, that's great! I don't feel great about removing those value checks, however. `callable()` takes less than a tenth of a microsecond on the machines I've tested. And builtins that take keys (like `sorted()`) raise errors on passing uncallables (as well as non-iterables), so I'd definitely like to retain that behavior.

Another weakness of this approach is that the resulting dictionary must be copied in order to return a plain dictionary. For small dictionaries (like even/odd in the example provided) this is fine, but it can be a problem for larger ones. Would a normal try/except be faster for this case? Thanks again!

coady:
### Type checks
Those TypeErrors will be raised organically though, as the the objects are iterated and called.  Builtins (like `sorted`) don't go out of their way to pre-check them.
```
In []: sorted([], key=0)
Out[]: []
```

### Performance
Yes, the smaller the buckets, the less the optimization.  But still the baseline is a relatively slow loop of O(V), whereas the additional copy is a fast inlined loop of O(K), where K <= V.  So it's not the size of the `dict` per se, but rather how little bucket-ing occurs.

One under-utilized feature of `defaultdict` is that if the `default_factory` is set to `None`, it behaves as a `dict` again.  That would be the fastest option of all.  Some comparison timings for buckets of size 2:
```
In []: %timeit bucketize(range(10 ** 6), (2).__rfloordiv__)
1 loop, best of 3: 428 ms per loop

In []: %timeit bucketize_default(range(10 ** 6), (2).__rfloordiv__)
1 loop, best of 3: 412 ms per loop

In []: %timeit bucketize_default_copy(range(10 ** 6), (2).__rfloordiv__)
1 loop, best of 3: 441 ms per loop
```

mahmoud:
But keeping things in perspective, an O(1) `callable()` call is really not going to make a difference, even for the microbenchmarks presented above. And as for `defaultdict`s, the main friction is that they don't `repr()` or `pprint()` the same. There may be other differences, but that alone is enough to make me uneasy about that most recent change.

I'm not sure if your experience has been otherwise, but to keep things in perspective, my usage of bucketize has been in not-so-performance-sensitive cases. It's a high-level function that saves a few lines, and has never been part of a hot loop or anything. As such, I care a lot more about usability, including good error messages. I'd like to work with those design directions in mind, until your (or others') experiences change them :)

PR #99, opened 2016-10-21
Add ioutils module

induane:
Spooled Temporary Files are file-like objects
that start out mapped to in-memory objects, but
automatically roll over to a temporary file once
they reach a certain (configurable) threshhold.
Unfortunately the built-in SpooledTemporaryFile
class in Python does not implement the exact API
that some common classes like StringIO do.
SpooledTemporaryFile also spools all of it's in-memory
files as cStringIO instances. cStringIO instances cannot
be deep-copied, and they don't work with the zip
library either. This along with the incompatible api makes
it useless for several use-cases.

To combat this but still gain the memory savings and
usefulness of a true spooled file-like-object, two custom
classes have been implemented which have a compatible API.

SpooledBytesIO is a spooled file-like-object that only
accepts bytes. On Python 2.x this means the 'str' type; on
Python 3.x this means the 'bytes' type. Bytes are written
in and retrieved exactly as given, but it will raise TypeErrors
if something other than bytes are written.

SpooledStringIO is a spooled file-like-object that only accepts
unicode values. On Python 2.x this means the 'unicode' type and
on Python 3.x this means the 'str' type. Values are accepted as
unicode and then coerced into utf-8 encoded bytes for storage. On
retrieval, the values are returned as unicode.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -46,3 +46,5 @@ nosetests.xml
 
 # Vim
 *.sw[op]
+
+.cache/
--- a/boltons/ioutils.py
+++ b/boltons/ioutils.py
@@ -293,20 +293,46 @@ class SpooledStringIO(SpooledIOBase):
         if self.tell() >= self._max_size:
             self.rollover()
 
+    def _traverse_codepoints(self, current_position, n):
+        """Traverse from current position to the right n codepoints"""
+        dest = current_position + n
+        while True:
+            if current_position == dest:
+                # By chance we've landed on the right position, break
+                break
+
+            # If the read would take us past the intended position then
+            # seek only enough to cover the offset
+            if current_position + READ_CHUNK_SIZE > dest:
+                self.read(dest - current_position)
+                break
+            else:
+                ret = self.read(READ_CHUNK_SIZE)
+
+            # Increment our current position
+            current_position += READ_CHUNK_SIZE
+
+            # If we kept reading but there was nothing here, break
+            # as we are at the end of the file
+            if not ret:
+                break
+
+        return dest
+
     def seek(self, pos, mode=0):
+        """Traverse from offset to the specified codepoint"""
+        # Seek to position from the start of the file
         if mode == os.SEEK_SET:
             self.buffer.seek(0)
-            self.buffer.read(pos)
-            return pos
+            return self._traverse_codepoints(0, pos)
+        # Seek to new position relative to current position
         elif mode == os.SEEK_CUR:
             start_pos = self.tell()
-            self.buffer.read(pos)
-            return start_pos + pos
+            return self._traverse_codepoints(self.tell(), pos)
         elif mode == os.SEEK_END:
             self.buffer.seek(0)
             dest_position = self.len - pos
-            self.buffer.read(dest_position)
-            return dest_position
+            return self._traverse_codepoints(0, dest_position)
         else:
             raise ValueError(
                 "Invalid whence ({0}, should be 0, 1, or 2)".format(mode)
--- a/boltons/ioutils.py
+++ b/boltons/ioutils.py
@@ -367,10 +367,6 @@ class SpooledStringIO(SpooledIOBase):
                 # needed to put us over the pos value.
                 pos_read = True
         else:
-            # We're done with the buffer at this point, so seek to the previous
-            # pos for happy return and error raising.
-            self.buffer.seek(pos)
-
             # We've read past the underlying seek value. We now iterate our
             # last retrieved chunk until the encoded character bytes plus the
             # current_pos add up to the pos. Anything over that means that our
@@ -379,14 +375,20 @@ class SpooledStringIO(SpooledIOBase):
                 enc_pos += 1
                 current_pos += len(char.encode('utf-8'))
 
-                if current_pos == pos:
+                if current_pos >= pos:
                     break
-                elif current_pos >= pos:
-                    raise IOError(
-                        "SpooledStringIO's underlying EncodedFile seeked to "
-                        "invalid code point!")
 
-        if current_pos != pos:
+        self.buffer.seek(pos)
+
+        if current_pos > pos:
+            # This means that we've read a character that started before the
+            # seek location, and ended after the seek location meaning that the
+            # cursor isn't aligned to the characters.
+            raise IOError(
+                "SpooledStringIO's underlying EncodedFile seeked to "
+                "invalid code point!")
+
+        if current_pos < pos:
             # If we made it this far and our current_pos doesn't equal our pos,
             # something has gone wrong with our flo.
             raise IOError(
@@ -394,7 +396,6 @@ class SpooledStringIO(SpooledIOBase):
                 "position. Likely data truncation while we were "
                 "reading it.")
 
-        self.buffer.seek(pos)
         return enc_pos
 
     @property
--- a/boltons/ioutils.py
+++ b/boltons/ioutils.py
@@ -9,8 +9,8 @@ are useful when dealing with input, output, and bytestreams in a variety of
 ways.
 """
 import os
+import sys
 import logging
-import six
 from abc import (
     ABCMeta,
     abstractmethod,
@@ -22,6 +22,27 @@ from codecs import EncodedFile
 from tempfile import TemporaryFile
 
 log = logging.getLogger(__name__)
+
+# Python2/3 compat
+if sys.version_info[0] == 3:
+    text_type = str
+    binary_type = bytes
+else:
+    text_type = unicode
+    binary_type = str
+
+
+def utf8_encode_if_unicode(string):
+    """
+    Given a string type, return it in bytes, encoding to UTF-8 if the string is
+    unicode.
+    """
+    if isinstance(string, text_type):
+        return string.encode('utf-8')
+    elif isinstance(string, binary_type):
+        return string
+    else:
+        raise TypeError("{} is not a unicode or str object".format(string))
 
 
 class SpooledIOBase(object):
@@ -173,7 +194,8 @@ class SpooledIOBase(object):
 
     def __eq__(self, other):
         if isinstance(other, SpooledIOBase):
-            return self.getvalue() == other.getvalue()
+            return (utf8_encode_if_unicode(self.getvalue()) ==
+                    utf8_encode_if_unicode(other.getvalue()))
         return False
 
     def __ne__(self, other):
@@ -189,21 +211,21 @@ class SpooledBytesIO(SpooledIOBase):
 
     Example::
 
-        >>> from lrmslib import ioutils
+        >>> from boltons import ioutils
         >>> with ioutils.SpooledBytesIO() as f:
         ...     f.write(b"Happy IO")
         ...     _ = f.seek(0)
-        ...     print(f.getvalue())
-        Happy IO
+        ...     isinstance(f.getvalue(), ioutils.binary_type)
+        True
     """
 
     def read(self, n=-1):
         return self.buffer.read(n)
 
     def write(self, s):
-        if not isinstance(s, six.binary_type):
+        if not isinstance(s, binary_type):
             raise TypeError("{0} expected, got {1}".format(
-                six.binary_type.__name__,
+                binary_type.__name__,
                 type(s).__name__
             ))
 
@@ -251,12 +273,12 @@ class SpooledStringIO(SpooledIOBase):
 
     Example::
 
-        >>> from lrmslib import ioutils
+        >>> from boltons import ioutils
         >>> with ioutils.SpooledStringIO() as f:
         ...     f.write(u"\u2014 Hey, an emdash!")
         ...     _ = f.seek(0)
-        ...     f.read()
-        u'\u2014 Hey, an emdash!'
+        ...     isinstance(f.read(), ioutils.text_type)
+        True
 
      """
 
@@ -264,9 +286,9 @@ class SpooledStringIO(SpooledIOBase):
         return self.buffer.read(n).decode('utf-8')
 
     def write(self, s):
-        if not isinstance(s, six.text_type):
+        if not isinstance(s, text_type):
             raise TypeError("{0} expected, got {1}".format(
-                six.text_type.__name__,
+                text_type.__name__,
                 type(s).__name__
             ))
 
--- a/boltons/ioutils.py
+++ b/boltons/ioutils.py
@@ -138,14 +138,13 @@ class SpooledIOBase(object):
     @property
     def len(self):
         """Determine the length of the file"""
-        # Calling .fileno() forces a rollover, so only access fileno if rolled
-        if self._rolled:
-            return os.fstat(self.fileno()).st_size
-
-        # Fallback if still in-memory file
         pos = self.tell()
-        self.seek(0, os.SEEK_END)
-        val = self.tell()
+        if self._rolled:
+            self.seek(0)
+            val = os.fstat(self.fileno()).st_size
+        else:
+            self.seek(0, os.SEEK_END)
+            val = self.tell()
         self.seek(pos)
         return val
 
--- a/boltons/ioutils.py
+++ b/boltons/ioutils.py
@@ -172,6 +172,9 @@ class SpooledIOBase(object):
         self.seek(pos)
         return val
 
+    def __len__(self):
+        return self.len
+
     def __iter__(self):
         yield self.readline()
 
--- a/boltons/ioutils.py
+++ b/boltons/ioutils.py
@@ -332,10 +332,10 @@ class SpooledStringIO(SpooledIOBase):
         self.seek(0)
         enc_pos = 0
         while True:
-            self.read(1)
-            enc_pos += 1
             if self.buffer.tell() == pos:
                 break
+            self.read(1)
+            enc_pos += 1
         self.buffer.seek(pos)
         return enc_pos
 
@@ -346,7 +346,7 @@ class SpooledStringIO(SpooledIOBase):
         self.seek(0)
         total = 0
         while True:
-            ret = self.read(32000)
+            ret = self.read(21333)
             if not ret:
                 break
             total += len(ret)
--- a/boltons/ioutils.py
+++ b/boltons/ioutils.py
@@ -148,7 +148,7 @@ class SpooledIOBase(object):
         # Emulate truncation to a particular location
         pos = self.tell()
         self.seek(size)
-        self.buffer.truncate(size)
+        self.buffer.truncate()
         if pos < size:
             self.seek(pos)
 
@@ -283,7 +283,7 @@ class SpooledStringIO(SpooledIOBase):
         super(SpooledStringIO, self).__init__(*args, **kwargs)
 
     def read(self, n=-1):
-        ret = self.buffer.read(n).decode('utf-8')
+        ret = self.buffer.reader.read(n, n)
         self._tell = self.tell() + len(ret)
         return ret
 
@@ -388,7 +388,7 @@ class SpooledStringIO(SpooledIOBase):
     def len(self):
         """Determine the number of codepoints in the file"""
         pos = self.buffer.tell()
-        self.seek(0)
+        self.buffer.seek(0)
         total = 0
         while True:
             ret = self.read(READ_CHUNK_SIZE)
--- a/boltons/ioutils.py
+++ b/boltons/ioutils.py
@@ -258,7 +258,7 @@ class SpooledBytesIO(SpooledIOBase):
         return val
 
     def tell(self):
-        return self._buffer.tell()
+        return self.buffer.tell()
 
 
 class SpooledStringIO(SpooledIOBase):
--- a/boltons/ioutils.py
+++ b/boltons/ioutils.py
@@ -232,11 +232,10 @@ class SpooledBytesIO(SpooledIOBase):
         """Roll the StringIO over to a TempFile"""
         if not self._rolled:
             tmp = TemporaryFile()
-            pos = self._file.tell()
-            self._buffer.seek(0)
-            tmp.write(self._buffer.read())
+            pos = self.buffer.tell()
+            tmp.write(self.buffer.getvalue())
             tmp.seek(pos)
-            self._buffer.close()
+            self.buffer.close()
             self._buffer = tmp
 
     @property
@@ -308,8 +307,7 @@ class SpooledStringIO(SpooledIOBase):
         if not self._rolled:
             tmp = EncodedFile(TemporaryFile(), data_encoding='utf-8')
             pos = self.buffer.tell()
-            self.buffer.seek(0)
-            tmp.write(self.buffer.read())
+            tmp.write(self.buffer.getvalue())
             tmp.seek(pos)
             self.buffer.close()
             self._buffer = tmp
--- a/boltons/ioutils.py
+++ b/boltons/ioutils.py
@@ -1,0 +1,304 @@
+# -*- coding: UTF-8 -*-
+
+# Coding decl above needed for rendering the emdash properly in the
+# documentation.
+
+"""
+Module ``ioutils`` implements a number of helper classes and functions which
+are useful when dealing with input, output, and bytestreams in a variety of
+ways.
+"""
+import os
+import logging
+import six
+from abc import (
+    ABCMeta,
+    abstractmethod,
+    abstractproperty,
+)
+from errno import EINVAL
+from io import BytesIO
+from codecs import EncodedFile
+from tempfile import TemporaryFile
+
+log = logging.getLogger(__name__)
+
+
+class SpooledIOBase(object):
+    """
+    The SpooledTempoaryFile class doesn't support a number of attributes and
+    methods that a StringIO instance does. This brings the api as close to
+    compatible as possible with StringIO so that it may be used as a near
+    drop-in replacement to save memory.
+
+    Another issue with SpooledTemporaryFile is that the spooled file is always
+    a cStringIO rather than a StringIO which causes issues with some of our
+    tools.
+    """
+    __metaclass__ = ABCMeta
+
+    def __init__(self, max_size=5000000):
+        self._max_size = max_size
+
+    @abstractmethod
+    def read(self, n=-1):
+        """Read n characters from the buffer"""
+        pass
+
+    @abstractmethod
+    def write(self, s):
+        """Write into the buffer"""
+        pass
+
+    @abstractmethod
+    def readline(self, length=None):
+        """Returns the next available line"""
+        pass
+
+    @abstractmethod
+    def readlines(self, sizehint=0):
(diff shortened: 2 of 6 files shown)


mahmoud:
Sounds intriguing! The code looks good, and I appreciate the thoroughness of your tests, as well. There are a couple architectural issues I'd like to resolve first:

1. ioutils usage of six is pretty minimal, can we bring that down to none? :) By definition, boltons [rely on no third-party packages](http://boltons.readthedocs.io/en/latest/architecture.html#integration). Looks like it's mostly for bytes types, so you can follow the same convention as [strutils](https://github.com/mahmoud/boltons/blob/master/boltons/strutils.py) and others.

2. Abstract Base Classes. I appreciate a consistent API as much as anyone, but in my experience, ABCs aren't super useful. They're checked at instantiation time, don't give great errors, and are generally kind of un-Pythonic in my opinion. Tests are the best way of enforcing these APIs :)

The first I'm pretty sure we need to resolve, but the second, I'm open to discussion if you feel strongly an ABC is best. Thanks for a great submission!

mahmoud:
Oh, one more thing, 99% sure we don't need that logger in there :)

induane:
Hi, thanks for such quick feedback. I realized the six issue after-the fact. I just finished patching it to remove the six dependency. I'll remove the unused logger as well.

My thinking here in using the Abstract base class was mostly due to experience I had using this library with others. There was a tendency to attempt to create instances of the base class (my original implementation defined the api and then raised NotImplementedErrors for the methods that the real classes would need to implement.

This turned out to be kind of a pain and I had to field a lot of 'WTF MATE?!' sorts of questions fairly frequently. I realize that that isn't exactly a technical reason for what amounts to a technical choice, but it seemed to be a good way to go after a few rounds of real world iteration.

induane:
This tox thing is pretty cool btw. Looks like it's not loving python 2.6 and pypy (two things I've never actually tested it with before).

mahmoud:
Hmm, yes, that's as good a justification as one can imagine for ABC.

I still see a couple of CI failures (2.6 and PyPy), let me know if you need with help with that. Ideally you could set tox up locally, but depending on your platform, it might be a headache.

Another thing that I think can be enhanced is if you added a couple of suitable examples. What comes to mind is like, file uploads or something? Not having used it myself, I'm probably not the best person to give examples of use :)

induane:
Yea - I'm not sure what is up with pypy there. That's a pretty different set of values it's getting in the unittests. If you have any idea about that I'd love a hand!

induane:
I've got tox running locally, it was a piece of cake to get setup. I'll see if I can figure out why pypy isn't doing what I expect and see about adding some extra documentation as well.

induane:
@mahmoud I think that's it; at least for now. The tests are fully passing and the documentation has been extended. Let me know if there is anything else you think is important!

Cheers!

 - Induane

boltons/ioutils.py
```diff
@@ -0,0 +1,322 @@
...
+        if self._rolled:
+            self.seek(0)
+            val = os.fstat(self.fileno()).st_size
+        else:
+            self.seek(0, os.SEEK_END)
+            val = self.tell()
+        self.seek(pos)
+        return val
+
+    def fileno(self):
```
markrwilliams: It seems weird that `fileno()` has a side effect, but the only other option is for this to change.  I think this is better.

boltons/ioutils.py
```diff
@@ -0,0 +1,322 @@
...
+
+    def __enter__(self):
+        return self
+
+    def __exit__(self, *args):
+        self._file.close()
+
+    def __eq__(self, other):
+        if isinstance(other, SpooledIOBase):
+            return (utf8_encode_if_unicode(self.getvalue()) ==
```
markrwilliams: I don't think you should be able to compare BytesIO backed instances to StringIO backed ones.   This should instead be
```python
if isinstance(other, self.__class__):
    return self.getvalue() == other.getvalue()
```

boltons/ioutils.py
```diff
@@ -0,0 +1,322 @@
...
+    def readlines(self, sizehint=0):
+        return self.buffer.readlines(sizehint)
+
+    def rollover(self):
+        """Roll the StringIO over to a TempFile"""
+        if not self._rolled:
+            tmp = TemporaryFile()
+            pos = self._file.tell()
+            self.__buffer.seek(0)
+            tmp.write(self.__buffer.read())
```
markrwilliams: Why not `self.__buffer.getvalue()`?  And also, don't use __!

boltons/ioutils.py
```diff
@@ -0,0 +1,322 @@
...
+    def readline(self, length=None):
+        return self.buffer.readline(length)
+
+    def readlines(self, sizehint=0):
+        return self.buffer.readlines(sizehint)
+
+    def rollover(self):
+        """Roll the StringIO over to a TempFile"""
+        if not self._rolled:
+            tmp = TemporaryFile()
```
markrwilliams: You can set the mode on `TemporaryFile`.  It defaults to binary on Python 3, but you can override it:

```python
TemporaryFile(mode='w+' + 'b' if not self._is_text else '')
```

boltons/ioutils.py
```diff
@@ -0,0 +1,322 @@
...
+
+    def readlines(self, sizehint=0):
+        return [x.decode('utf-8') for x in self.buffer.readlines(sizehint)]
+
+    @property
+    def buffer(self):
+        try:
+            return self.__buffer
+        except AttributeError:
+            self.__buffer = EncodedFile(BytesIO(), data_encoding='utf-8')
```
markrwilliams: You don't need to do this if you set the mode on the `TemporaryFile` in the base implementation.

tests/test_ioutils.py
```diff
@@ -0,0 +1,275 @@
+import os
```
markrwilliams: These tests are good!

boltons/ioutils.py
mahmoud: yeah, i was gonna shorten the `__` to a `_` during my merge. :)

mahmoud:
I think I largely agree with Mark's comments. @induane, I'm happy to follow through with them on merge, but you went through the trouble of getting tox working, so if you'd like to implement them I'd very much appreciate it, too :)

boltons/ioutils.py
```diff
@@ -0,0 +1,322 @@
...
+    """
+    __metaclass__ = ABCMeta
+
+    def __init__(self, max_size=5000000):
+        self._max_size = max_size
+
+    @abstractmethod
+    def read(self, n=-1):
+        """Read n characters from the buffer"""
+        pass
```
induane: I know, but I find that a lot of people don't get that so I usually go the pass route for clarity.

boltons/ioutils.py
```diff
@@ -0,0 +1,322 @@
...
+    @property
+    def pos(self):
+        return self.tell()
+
+    @property
+    def buf(self):
+        return self.getvalue()
+
+    @property
+    def len(self):
```
induane: That could be implemented as well, but the goal here was to match the api of the StringIO module, hence this property. Without it, SpooledBytesIO and SpooledStringIO wouldn't be possible to use as drop-in replacements.

boltons/ioutils.py
induane: I don't mind changing it though!

.gitignore
```diff
@@ -46,3 +46,5 @@ nosetests.xml
 
 # Vim
 *.sw[op]
+
+.cache/
```
induane: The tests leave .cache folder behind, and I kept accidentally trying to commit it because I get lazy with git add -A

boltons/ioutils.py
induane: I thought it was weird myself. It behaves differently on pypy vs any of the cpython implementations.

boltons/ioutils.py
mahmoud: hehe, pass is probably clearer :)

boltons/ioutils.py
induane: Yea it may not make a lot of sense - I'm not sure myself. I figured it could be useful in a few instances, but I don't have strong feeling about this either way so I'll defer to ya'll.

boltons/ioutils.py
induane: I'm not 100% clear on what you're suggesting here. Do you mean I should match the modes between python versions by specifying it directly?

boltons/ioutils.py
induane: Hey, you're knocking my bad habits!!

You're right though, getvalue() probably makes more sense.

boltons/ioutils.py
induane: Which thing don't i need to do?

tests/test_ioutils.py
induane: Thanks!

boltons/ioutils.py
mahmoud: Yes, but this seems like a good default. The PyPy behavior is kinda curious. Best way to ask those folks is to jump on IRC probably.

boltons/ioutils.py
induane: I agree - implementing __len__ would be trivial too, so there aren't many reasons to leave it out. And yea - I see a lot of StringIO used, particularly in proprietary code. This often leads to horrible memory issues over time, so having a more memory efficient drop-in replacement is quite handy when you need to reduce memory overhead.

boltons/ioutils.py
markrwilliams: Since `pass` is executable code, Coverage will report these methods as uncovered even though they can't be run.  Docstrings, however, are not executable code, so they don't need to be covered.  The resulting Coverage report will be more accurate.

boltons/ioutils.py
markrwilliams: @induane What behavior do you see under PyPy?

boltons/ioutils.py
markrwilliams: @induane What I'd meant was that you could have `TemporaryFile` handle encoding and decoding for you.  But this will only work on Python 3.

However, I've changed my mind about this approach!  I'll write more about it in a comment on `EncodedFile`.

boltons/ioutils.py
markrwilliams: @induane There'd be no need for `EncodedFile` if `TemporaryFile` could be made to handle encoding.  But it that relies on behavior in Python 3 so it's irrelevant.

It turns out, though, that `EncodedFile` is useless here.  That's because it translates a byte string in one encoding to a bye string in another encoding, and stores the latter in a file.  So this is an error:

```python
codecs.EncodedFile(io.BytesIO(), data_encoding='utf-8').write(u'\u2022')
```

Because encoded file wants bytes as input.

I think what you actually want is `codecs.StreamReaderWriter`.  That's what `codecs.open` uses.  Here's how you might replace `EncodedFile` with it:

```python
self._mem_buf = BytesIO()
codec_info = codecs.lookup('utf-8')
self._buffer = codecs.StreamReaderWriter(self._mem_buf,
                                         codecs_info.reader,
                                         codecs_info.writer)
```

With that change you won't need to utf-8 encode any of the reads or the writes to `self.buffer`.  However, you'll need to query the in-memory buffer to find out the length of the encoded data before a rollover.  e.g.:

```python
@property
def len(self):
    if self._rolled:
        ...
    else:
        return len(self._mem_buf.getvalue())
```

You'll also need to implement `tell()` in the same way, as `StreamReaderWriter` has a `seek` but no `tell` method.

If all that is too much work, you can just get rid of `EncodedFile` entirely.  The advantage there would be that the `rollover` implementation between the two instances would become identical.

One last comment:  it appears that you always want the number of bytes of *encoded* data, not the number of code points, so that writing `u'\u2022'` will increase the length by 3, not 1.  If it were possible to return the number of code points instead (like `io.StringIO` would), then you could avoid encoding when writing to an in memory buffer by using `io.StringIO`.  That doesn't appear to be the desired API, though.

mahmoud: Haha short and sweet (as to be expected with Python and encodings!). I think I understand now, at least. How about you, @induane, are you comfortable with these suggestions?

boltons/ioutils.py
induane: I'll remove it; hopefully most of the people reading through it will know what's up.

boltons/ioutils.py
induane: @markrwilliams  calling os.fstat(filno_here).st_size behaves differently on pypy. In all other implementations, that call returns the file size. On pypy it returns the number of bytes from the current cursor position within the file. Thus if you don't seek back to zero, st_size reports an incorrect value.

boltons/ioutils.py
induane: @markrwilliams  In thinking about this a bit more, one can certainly expect someone to have written ascii characters only into either type of class. On Python u"hi" == b"hi" turns out to be True. Is there a compelling reason for similar behaviour not to be carried forth into these classes? After all these are basically file like object classes that each represent either bytes or text strings; the current behaviour is analogical to the text string builtins.

boltons/ioutils.py
induane: @mahmoud  @markrwilliams 

It's hard to say. I tried for quite a while without success to get the StreamReader/Writer working but with little success. I don't know it as well though so it's probable I got something messed up due to lack of familiarity. Well it isn't that I didn't get it working, it's that it broke the tests in horrible ways and I couldn't figure out [quickly] how to go about rewriting things so that they worked as expected.

Also @markrwilliams  with regards to the len property, you're right. There I'm returning the actual file size in one form or another. The variable byte width of utf-8 of course does mean exactly what you imply. I'm wondering though if this really is the best approach.

I've gone the route of using len to indicate actual size, but implemented seek, tell, etc... in a way that moves across the code points rather than individual bytes (so that you couldn't seek to a point in the middle of a character so to speak).

Maybe the smart approach here is to make it work as you suggest as that more closely matches what StringIO does.

induane: sorry tell isn't implemented that way  - it gives you the real position. Thus you get something akin to:

```
>>> flo.write(u'\u2014')
>>> flo.seek(0)
>>> flo.read(1)
—
>>> flo.tell()
3L
```

boltons/ioutils.py
markrwilliams: On Python 2, `u"hi" == b"hi"` because the byte string (`b"hi"`) is implicitly decoded into a `unicode` object using the system's default encoding, which is `ascii`.  That means this will be `False`:

```python
u'u\2022' == b'\xe2\x80\xa2'
```

and produce a warning about how the comparison fails because both arguments could not be converted to `unicode`.

Your implementation would have `SpooledBytesIO(u'\u2022') == SpooledStringIO('\xe2\x80\xa2') return `True`, breaking the illusion of consistency with Python 2's string types.  
Also, comparing a `SpooledBytesIO` object containing zipped data to a `SpooledStringIO` object containing text could raise a `UnicodeDecodeError`, when I'd expect it to return `False`.

And finally, implicit decoding is hard enough to remember in all circumstances that it should be considered an anti-pattern.

induane:
@mahmoud @markrwilliams  I added these comments previously as a reply to a comment and probably stuck them in a review or something because I wasn't paying attention. In any case, I'm throwing it back here on the main comments thread so it doesn't get lost.

As for being fine with the StreamReader/Write changes, it's hard to say. I tried for quite a while without success to get that to work with little success. I don't know it as well though so it's probable I got something messed up due to lack of familiarity rather than there being a legitimate problem with the suggested approach. Actually, I kind of got it working but in a way that it broke the tests in horrible ways and I couldn't figure out [quickly] how to go about rewriting things so that they worked as expected.

Also @markrwilliams with regards to the len property, you're right. There I'm returning the actual file size in one form or another. The variable byte width of utf-8 of course does mean exactly what you imply. I'm wondering though if this really is the best approach. I've gone the route of using len to indicate actual size, but implemented seek, read, etc... in a way that moves across the code points rather than individual bytes (so that you couldn't seek to a point in the middle of a character so to speak). Maybe the smart approach here is to make it work as you suggest as that more closely matches what StringIO does, but I'm not actually sure how to make that happen.

boltons/ioutils.py
induane: Yea fair enough. That's a pretty trivial change too so I'll go ahead and update the patchset. Thanks for responding!

boltons/ioutils.py
induane: @markrwilliams  I don't think we want seek and read to act on bytes (that's really what EncodedFile actually buys). If you drop out that, then seek and read don't work the way they do on StringIO and that prevents this from being a useful drop in replacement. I don't know exactly how to make tell() not provide info about what byte you're on (and maybe it shouldn't).

Really though I think in retrospect, len on SpooledStringIO should report the number of codepoints and not the number of bytes (otherwise again we're off of the StringIO api). One of the reasons that these classes are useful is that they provide a consistent API for different kinds of file and file-like objects. 

I have an idea for making len work as expected. I'll think about tell a bit more.

induane: Seek is already implemented to operate on codepoints. I updated the pull request an implemented len and tell as well, but tell is pretty inefficient. I am looking at doing it in larger chunks when possible, something like 21333 which even if all 21333 characters are 3 btyes long, falls under 64000 byes, so it'll nicely fit into the L2 cache of most processors without any problems.

I'm thinking in terms of the tell implementation, a better option than what I did in my updated pull request would be to check the position (in bytes) then seek in chunks, recording each along the way, until the chunk read takes you past the original tell() point. At that point you could back up and read in smaller increments until you hit the byte position. That would be relatively efficient.

induane: @markrwilliams  Sorry for the huge pile-o-comments that have stacked up. I really appreciate all the back-and-forth.

I've written an improved version of codepoint supported tell and pushed that (as described in my previous message). Would you have a chance to look at the latest code in the pull request?

mahmoud:
Just chiming in to say I'm loving the discussion. I was planning a boltons release this last weekend, but I'm glad I held off. This is a very challenging area and it's only right to give it some thought. Let me know if there are any merges/branches I can make that'd enable even more collaboration :)

induane:
@mahmoud  I'm having a lot of fun with it too; the back and forth forced me to think about a few of the implementation specifics in greater detail and I think the resulting changes were really useful. I've already pulled them back into the version I'm currently using for projects.

induane:
@markrwilliams  @mahmoud  Alright - at this point I'm happy as a lark with the implementation I think. Let me know if there is anything else I need to clean up.

mahmoud:
I've been quite happy for a while! Unless Mark or you tell me otherwise, I'll be merging this tonight and we'll handle any further improvements on a separate PR. :)

markrwilliams:
@induane I'm really glad you enjoyed the discussion and found it productive :)  I'm going to take one more pass over the PR and submit a new CR.

boltons/ioutils.py
```diff
@@ -0,0 +1,439 @@
...
+            current_position += READ_CHUNK_SIZE
+
+            # If we kept reading but there was nothing here, break
+            # as we are at the end of the file
+            if not ret:
+                break
+
+        return dest
+
+    def seek(self, pos, mode=0):
```
markrwilliams: I wrote a small [Hypotheis](http://hypothesis.readthedocs.io/en/latest/) test case for `seek` with `SEEK_SET`:

```python
import pytest
import io

from boltons import ioutils
from hypothesis import strategies as st, given


def make_spooled_stringio():
    return ioutils.SpooledStringIO(32)


@given(text=st.text(),
       data=st.data())
@pytest.mark.parametrize('make_sio', [
    io.StringIO,
    make_spooled_stringio,
])
def test_seek(text, data, make_sio):
    where = data.draw(st.integers(min_value=0, max_value=len(text)))
    s = make_sio()
    s.write(text)
    s.seek(where)
    assert s.read() == text[where:]
```

The invariant under test is that seeking forward from the beginning any amount between 0 and the length of the string and then reading the remainder should return a string equal to the slice beginning there.

All Hypothesis test data passes with `io.StringIO`, but it fails with `SpooledStringIO` and the following inputs:

```python
sio = SpooledStringIO(32)
sio.write(u'\x8000')
sio.seek(2)
sio.read()  # prints u'', not u'0'
```

This appears to be unexpected behavior courtesy of `EncodedFile`

```
>>> c = EncodedFile(io.BytesIO(), data_encoding='utf-8')
>>> c.write(u'\x8000'.encode('utf-8'))
>>> c.seek(0)
0L
>>> len(c.read(2).decode('utf-8'))
3
````

That is, a request to read 2 code points actually returns 3!

It's certainly true that *your* code is on point!  But given the difficulty presented by `EncodingFile`, It occurs to me that it may be worthwhile to take a different approach.  What if you used `utf-32` instead of `utf-8`?  Because UTF-32 is a fixed-width encoding, you wouldn't need `EncodedFile`, but instead could just do modulo 4 arithmetic for seeks and tells.  It's worth noting that CPython itself uses one of several fixed-width representations of Unicode strings to avoid the hairy code you've successfully written and also for the sake of greater efficiency; with a fixed width encoding you can seek backwards from the end without having to go forward from the beginning!

Comment (mahmoud) [inline boltons/ioutils.py]: Ah, you know the code is :ok_hand: when the decorators are as long as the function body. But at least it does its job, unlike some of these shadowy corners of the stdlib. :/

Yeah, after a bit of thought, the easy layout of UTF-32 does seem better. Despite all the great work so far, the analogy of UCS4 in-memory rings true, since we're use a temporary file like a memory spillover.

Personally I'm more psyched about the bytes version, but it would be great if the text version was comparatively reliable. How does that sound, @induane?

File: boltons/ioutils.py [hunk 11]
```diff
@@ -0,0 +1,439 @@
...
+                # By chance, our enc_pos is some multiple of READ_CHUNK_SIZE.
+                # We're done!
+                break
+
+            chunk = self.read(READ_CHUNK_SIZE)
+            chunk_bytes = len(chunk.encode('utf-8'))
+
+            if current_pos + chunk_bytes < pos:
+                current_pos += chunk_bytes
+                enc_pos += len(chunk)
```
induane: Not if you've reached the end of the file.

boltons/ioutils.py
```diff
@@ -0,0 +1,439 @@
...
+        return enc_pos
+
+    @property
+    def len(self):
+        """Determine the number of codepoints in the file"""
+        pos = self.buffer.tell()
+        self.seek(0)
+        total = 0
+        while True:
+            ret = self.read(21333)
```
induane: Def!

boltons/ioutils.py
coleifer: `u'\x8000'` doesn't mean anything special. You get unicode code point 128 (\x80) and two zeroes, hence len == 3.

```python
In [1]: u'\x8000' == unichr(128) + '00'
Out[1]: True
```

coleifer: u'\x8000' doesn't mean anything special. You get unicode code point 128 (\x80) and two zeroes, hence len == 3.

In [1]: u'\x8000' == unichr(128) + '00'
Out[1]: True

induane: @coleifer  The problem is that the request was for 2 characters.

markrwilliams: Right.  The problem is that the read was asked for 2 code points but returned 3.

With `io.StringIO`:

```python
>>> len(io.StringIO(u'\x8000').read(2))
2
```

Again, with `EncodedFile`:

```python
>>> len(codecs.EncodedFile(io.BytesIO(u'\x8000'.encode('utf-8')), data_encoding='utf-8').read(2).decode('utf-8'))
3
```

In this case, there *is* something special about u'\x80', because ASCII encodable code points work as expected:

```python
>>> len(codecs.EncodedFile(io.BytesIO(u'abc'.encode('utf-8')), data_encoding='utf-8').read(2).decode('utf-8'))
2
```

induane:
@mahmoud  @markrwilliams Okay - so I thought about it a lot more. I've fixed the implementation so that it will pass @markrwilliams devious and horrifying (to me) tests. I've included an implementation of his test in the test suite, just for defensive purposes.

I nearly went ahead and with the utf-32 route - but I got partway, realized that the byte markers created an annoying offset that was making the implementation kind of annoying; at least to me it was about as annoying as utf-8. I suspect that the main use cases for SpooledStringIO are to be replacing StringIO, and StringIO with unicode characters tends to come into play when people need a file like object interface for something, but can't control whether or not they get random unicode characters. Dropping this in can have memory savings; with utf-32 I'd bet that most of the time smaller sets of data would be inflated by 4x as most of the characters are likely to be ascii. That made the nice variable byte width of utf-8 so enticing that I just couldn't resist the savings. 

The implementation of seek and len are probably the only parts at this point that takes any measurable performance hit from implementing seek on top of reads. Most of the time though I don't think that people will be doing lots of arbitrary seeking (except maybe back to zero).

mahmoud:
Looks pretty clean to me. Great job on getting EncodedFile working (and the respective test). I'm going to go ahead and merge, and future versions based on UTF-32 are of course welcome. Thanks again everyone for all your great comments and contributions!

markrwilliams:
@induane Sorry I didn't get a chance to comment before this got merged.  Here are a couple of additional Hypothesis tests I wrote:

```python
from boltons import ioutils
from hypothesis import strategies as st, given
import io
import os
import pytest


def make_spooled_stringio():
    return ioutils.SpooledStringIO(32)


def draw_where(text, data):
    return data.draw(st.integers(min_value=0, max_value=len(text)))


@given(text=st.text(),
       data=st.data())
@pytest.mark.parametrize('make_sio', [
    io.StringIO,
    make_spooled_stringio,
])
def test_seek(text, data, make_sio):
    where = draw_where(text, data)
    s = make_sio()
    s.write(text)
    s.seek(where)
    assert s.read() == text[where:]


@given(text=st.text(),
       data=st.data())
def test_seek_end(text, data):
    where = draw_where(text, data)
    s = make_spooled_stringio()
    s.write(text)
    s.seek(where, os.SEEK_END)
    assert s.read() == text[len(text) - where:]


@given(text=st.text(),
       data=st.data())
def test_seek_cur(text, data):
    where = draw_where(text, data)
    s = make_spooled_stringio()
    s.write(text + text)
    s.seek(len(text))
    s.seek(where, os.SEEK_CUR)
    assert s.read() == text[where:]


@given(text=st.text())
@pytest.mark.parametrize('make_sio', [
    io.StringIO,
    make_spooled_stringio,
])
def test_tell(text, make_sio):
    s = make_sio()
    s.write(text)
    assert s.tell() == len(text)
```

You might be pleased to know that all tests pass :) 

Thanks so much for all the work you did on this.  Your code is great!  I'm especially impressed with the solution you found to the `u'\x80'` problem.

I'm really looking forward to your next PR.

mahmoud:
@induane thought you might like to know this is now pip installable in boltons 16.5.1. Thanks again for your patience and hard work!

induane:
@mahmoud that's awesome to hear, thanks for letting me know!

PR #101, opened 2016-10-28
Rollover to file before writing overflow value

induane:
If the write operation is going to push a
SpooledBytesIO instance beyond it's max size, dispatch
a rollover to disk before performing the write. This
prevents duplicating a giant slug of data into memory
if a write of lots of data is attempted. This optimization
was made already for the SpooledStringIO write method
but was omitted by mistake from the SpooledBytesIO
implementation.


mahmoud:
Seems uncontroversial, the code looks right, tests pass, all good!

PR #103, opened 2016-11-08
Fix ioutils documentation missing from index.rst

induane:


mahmoud:
Speak of the devil, I noticed this about 10 minutes before this commit. Thanks!

PR #104, opened 2016-11-14
Implement __nonzero__ for SpooledBytesIO, SpooledStringIO

induane:
An empty StringIO instance still has a "truthy" value. In
order to user SpooledBytesIO with the ZipFile library, the
object instance needs to be considered truthy. This patch
implements __nonzero__ so that the truthyness does not fall
back on the underlying __len__ implementation.

Added a test that verifies a SpooledBytesIO is compatible
with the ZipFile library.

### DIFF
--- a/boltons/ioutils.py
+++ b/boltons/ioutils.py
@@ -180,11 +180,10 @@ class SpooledIOBase(object):
     def __ne__(self, other):
         return not self.__eq__(other)
 
-    def __nonzero__(self):
-        return self.__bool__()
-
     def __bool__(self):
         return True
+
+    __nonzero__=__bool__
 
 
 class SpooledBytesIO(SpooledIOBase):
--- a/boltons/ioutils.py
+++ b/boltons/ioutils.py
@@ -180,6 +180,9 @@ class SpooledIOBase(object):
     def __ne__(self, other):
         return not self.__eq__(other)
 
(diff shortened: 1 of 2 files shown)


mahmoud:
haha, I can't be the only one to think that it's strange that `bool(io.BytesIO()) == True`, but, um, I'm not in charge of that project, and all I can say is that this change seems to make sense, given certain uncontrollabe contexts. Merged. Thanks!

induane:
@mahmoud You're definitely not the only one to think it's weird. It kind of boggled my mind too. I guess even an empty file is considered a thing, and a StringIO is a file like object, as is BytesIO, etc..

If you dig into the ZipFile code: https://svn.python.org/projects/python/trunk/Lib/zipfile.py you'll find this delightful chunk of code in it's open method:

```python
def open(self, name, mode="r", pwd=None):
    """Return file-like object for 'name'."""
    if mode not in ("r", "U", "rU"):
        raise RuntimeError, 'open() requires mode "r", "U", or "rU"'
    if not self.fp:
        raise RuntimeError, \
              "Attempt to read ZIP archive that was already closed"
```

So the if you pass in a flo, it's truthy value is checked. I think that should probably be changed to `if self.fp.closed()` or something but that might be a moot point now, that library has been around for ages.

PR #105, opened 2016-11-16
Add FrozenDict type to dictutils

induane:
This is another tool I have built a number of times in various permutations; 
I don't know if it's all that interesting but I've found it quite useful in a number
of situations including immutable constant maps without heinous amounts of
namedtuple boilerplate, and some cases with threading. 

A FrozenDict is a dictionary whose values are
set during instance creation and are fixed from
that point on. Setting and altering values is
not allowed. This can be useful in a number of
scenarios, including setting up mapping constants
without worrying that the values will get mutated
by a misbehaving function.

There are a lot of online recipes for creating a
FrozenDict class, but most still rely on a real
dictionary under the hood for storage. They also
tend to be extra weighty because they have an
underlying object dict.

One common solution is to use a named tuple, but this
requires setting up boilerplate for every type or
relying on factory functions. The FrozenDict utilizes
a named tuple for storage and then is further made
lighter by utilizing __slots__ to eliminate the
underlying object dict.

### DIFF
--- a/boltons/dictutils.py
+++ b/boltons/dictutils.py
@@ -34,7 +34,7 @@ thanks to `Mark Williams`_ for all his help.
 .. _Mark Williams: https://github.com/markrwilliams
 """
 
-from collections import KeysView, ValuesView, ItemsView, namedtuple, Mapping
+from collections import KeysView, ValuesView, ItemsView
 
 try:
     from itertools import izip_longest
@@ -694,7 +694,56 @@ class FastIterOrderedMultiDict(OrderedMultiDict):
             curr = curr[PREV]
 
 
-class FrozenDict(Mapping):
+class SentinelInt(int):
+    pass
+
+
+class FrozenDictBase(tuple):
+
+    def __getitem__(self, key):
+        if isinstance(key, SentinelInt):
+            return tuple.__getitem__(self, key)
+        else:
+            try:
+                return getattr(self, str(key.__hash__()))()
+            except AttributeError:
+                raise KeyError(key)
+            except IndexError:
+                raise KeyError(key)
+
+    def __iter__(self):
+        idx = 0
+        while True:
+            try:
+                yield self[SentinelInt(idx)][0]
+            except IndexError:
+                break
+            idx += 1
+
+    def iteritems(self):
+        """Iterates, returning key, value pairs as a tuple"""
+        idx = 0
+        while True:
+            try:
+                yield self[SentinelInt(idx)][0], self[SentinelInt(idx)][1]
+            except IndexError:
+                break
+            idx += 1
+
+    def get(self, key, default=None):
+        try:
+            return self[key]
+        except KeyError:
+            return default
+
+    def keys(self):
+        return [x[0] for x in self]
+
+    def __contains__(self, value):
+        return hasattr(self, str(value.__hash__()))
+
+
+def FrozenDict(dict_map):
     """A FrozenDict is a dictionary whose values are set during instance
     creation and are fixed from that point on. Setting and altering values is
     not allowed. This can be useful in a number of scenarios, including setting
@@ -707,49 +756,29 @@ class FrozenDict(Mapping):
 
     One common solution is to use a named tuple, but this requires setting up
     boilerplate for every type or relying on factory functions. The FrozenDict
-    utilizes a named tuple for storage and then is further made lighter by
-    utilizing __slots__ to eliminate the underlying object dict.
+    auto-generates a special subclass of Tuple that includes fast attribute
+    lookups and full immutability
 
     Declaration of a FrozenDict requires passing all desired values into the
     constructor:
 
-    >>> fd = FrozenDict(a=1, b=2, c=3)
+    >>> fd = FrozenDict({"a": 1, "b": 2, "c": 3})
     >>> fd.get('a')
     1
     >>> fd['b']
     2
     """
+    fields = {}
+    base = []
+    for idx, val in enumerate(dict_map.items()):
+        getter = lambda self, k=SentinelInt(idx): self[k][1]
+        fields[str(val[0].__hash__())] = getter
+        base.append((val[0], val[1]))
+    f = type('ImmutableDict', (FrozenDictBase, ), fields)
+    return f(base)
 
-    __slots__ = ('_storage', '_hash_val')
 
-    def __init__(self, **kwargs):
-        self._storage = namedtuple("FrozenDictStore",
-                                   " ".join(kwargs.keys()))(**kwargs)
+frozendict = FrozenDict
 
-    def __iter__(self):
-        return iter(self._storage._fields)
-
-    def iteritems(self):
-        """Iterates, returning key, value pairs as a tuple"""
-        for key in self._storage._fields:
-            yield key, getattr(self._storage, key)
-
-    def __len__(self):
-        return len(self._storage)
-
-    def __getitem__(self, key):
-        try:
-            return getattr(self._storage, key)
-        except AttributeError:
-            raise KeyError(key)
-
-    def __hash__(self):
-        try:
-            return self._hash_val
-        except AttributeError:
-            self._hash_val = 0
-            for tuple_val in self.iteritems():
-                self._hash_val ^= hash(tuple_val)
-        return self._hash_val
 
 # end dictutils.py
--- a/boltons/dictutils.py
+++ b/boltons/dictutils.py
@@ -720,6 +720,10 @@ class FrozenDictBase(tuple):
                 break
             idx += 1
 
+    def items(self):
+        """Returns a list of tuples of data"""
+        return [(x, self[x]) for x in self]
+
     def iteritems(self):
         """Iterates, returning key, value pairs as a tuple"""
         idx = 0
@@ -745,6 +749,11 @@ class FrozenDictBase(tuple):
     def __contains__(self, value):
         return hasattr(self, str(value.__hash__()))
 
+    def __reduce__(self):
+        state = {}
+        for k, v in self.iteritems():
+            state[k] = v
+        return (FrozenDict, (state, ))
 
 def FrozenDict(dict_map):
     """A FrozenDict is a dictionary whose values are set during instance
--- a/boltons/dictutils.py
+++ b/boltons/dictutils.py
@@ -34,7 +34,7 @@ thanks to `Mark Williams`_ for all his help.
 .. _Mark Williams: https://github.com/markrwilliams
 """
 
-from collections import KeysView, ValuesView, ItemsView
+from collections import KeysView, ValuesView, ItemsView, namedtuple, Mapping
 
 try:
     from itertools import izip_longest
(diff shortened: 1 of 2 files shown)


induane:
For reference:

https://www.python.org/dev/peps/pep-0416/

The pep details some constraints:

- frozendict has to implement the Mapping abstract base class
- frozendict keys and values can be unorderable
- a frozendict is hashable if all keys and values are hashable
- frozendict hash does not depend on the items creation order

Most implementations referenced in the Pep don't match the constraints fully; this may need touched up but I think it does (although it's not implemented in C which would be even cooler).

mahmoud:
Hey again! This is definitely something I've seen coming up in various projects. At first blush, I like the brevity of the implementation, and it seems to get the job done. I'll take it for a spin and get back to you soon. Thanks for another great submission!

sigmavirus24:
So the only other `frozen*` in Python is `frozenset` which is a builtin. While I understand that there's `OrderedDict`, I think `frozendict` would be better. I suspect the reason for `OrderedDict`'s casing is that it was originally implemented in pure Python and not in C, so if you wanted to use that as a distinguishing factor, that could work. That said, `namedtuple` is "implemented" in pure python, so *shrug*.

kurtbrose:
The number of implementations linked to from the PEP is really impressive.  I think this is clearly an itch a lot of people are having that could benefit from a good solution. 

This implementation is very elegant, the code is short clear and readable.  My only concern would be that namedtuple() constructor is a bit slow since it is calling compile().

```
>>> timeit.timeit("collections.namedtuple('a', 'a b c d e f g h')", "import collections", number=1000)
0.46814227974070377
>>>
>>> timeit.timeit("frozenset('abcdefgh')", number=1000)
0.0007556269890756084```

I think this means a namedtuple based frozendict would be ~500x slower than a frozenset.

An alternative implementation might be something like this:

```python
class ordereddict(object):
   __slots__ = 'items', 'lookup'

   def __init__(self, map):
       self.items = tuple(map.items())
       self.lookup = dict(map)

   def __hash__(self):   return hash(self.items)
   def __getitem__(self, k): return self.lookup[k]
```

Maybe some of the other ones in Werkzeug and SQLAlchemy already do something like this?

mahmoud:
@sigmavirus24 yeah I could see it going either way, but I do slightly prefer the lowercase. It's meant to be a primitive, like frozenset. That said, other folks I've shopped it around to have pointed at OrderedDict and UserDict (and even boltons's own OrderedMultiDict), so I think the TitleCase may have to stay. Tell ya what though, I'd totally be down with adding a lowercased alias, what do you think of that?

@kurtbrose I think that other approaches just use method wrappers that raise errors. The compile overhead observation is a very good one, and the frozenset approach is a great idea! @induane what do you think?

kurtbrose:
Hm... the implementation of hash() also is going to be low performance -- it's going to go around iteritems() for each element in the dict, calling getattr() then xor'ing together python integers.  If possible it would be best if all the methods were simple dispatches to built-in objects.  (Since this is frozen there are no thread concurrency issues.)

Another possible approach, this one trades off log-n lookups for better memory usage:

```python
class frozendict(tuple):
   def __new__(self, map):
      tuple.__new__(self, sorted(map.items()))

   def __getitem__(self, key):
      ix = bisect.bisect_left(self)
      if self[ix + 1][0] == key:
         return self[ix + 1][1]
      raise KeyError()
```

induane:
I really like the concept of python enforce immutability; it's one of the things that I was thinking about after submitting the pull request and hadn't really conjured a good way of doing it. The version in the PR is pretty much the one I've had floating around in various incarnations for quite a while and so giving it some real thought is worth it.

I like @kurtbrose's base implementation that subclasses tuple, but it will make the implementation a bit more involved to really round out the mapping interface which I had previously managed to do by subclassing Mapping from collections.

The lookups bit though seems like the more important attribute; though I guess it depends on intended usage. On of the advantages of a hashmap is the lookup speed; so being as close to performant there as possible seems prudent.

I'd like to toy with this a bit I think; in thinking about it, it occurs to me another limitation of the initial implementation I posted is that it's not possible to make any hashable object into a key.

induane:
Also @kurtbrose the hash implementation might be trouble on excessively large numbers of dictionary keys but it'll only ever be calculated once per object instance so I'd suspect it might not be the worst thing ever. It's not always going to be calculated either.

One of the reasons to implement hash is so that an object can be used in a hashmap. If the frozendict (FrozenDict?) is hashable using an underlying hash then it is possible that both a frozendict and an identical object (tuple, whatever is used behind the scenes) could end up with an identical hash value which in turn could cause key collisions. That's probably such an obscure scenario that it's not a big deal, but possibly worth considering as well.

kurtbrose:
@induane 100% agree on __hash__ being critical to the usefulness of frozendict!

That's a good point about unintentional object equality which I hadn't considered.  A good API should definitely take that into account.  (In my opinion True/1/enum.IntEnum overwriting each other in dicts is an oversight in stdlib.)

One way that could be handled would be a custom __eq__ that takes into account types.  (Two objects with the same __hash__ but that test as not equal will not alias each other in a dict).

```python
>>> class T(tuple): pass
... 
>>> T([1,2]) == (1,2)
True
>>> class T(tuple):
...    def __eq__(self, other):
...       return type(self) == type(other) and tuple.__eq__(self, other)
... 
>>> T([1,2]) == (1,2)
False
>>> T([1,2]) == T((1,2))
True
>>> {T( (1,2) ): 'cat', (1,2): 'dog'}  # objects don't alias each other
{(1, 2): 'cat', (1, 2): 'dog'}
```

Also, I totally missed the `self._hash_val` memoization, that does help with the computational expense.

kurtbrose:
BTW, there's a hole you can drive a truck through in my idea of `sort` + `bisect`, which is that `sort()` doesn't work in many cases: http://stackoverflow.com/questions/26575183/how-can-i-get-2-x-like-sorting-behaviour-in-python-3-x

For example, complex numbers are perfectly valid as dictionary keys but cannot be sorted:
```python
>>> 1j > 1j
  File "<stdin>", line 1, in <module>
TypeError: no ordering relation is defined for complex numbers
>>> {1j: 'cat'}
{1j: 'cat'}
```

I'm just hoping O(n) `__getitem__` can be avoided somehow.  Ideally what you'd want is a way to ask a set / frozen set "at what index did you store this object?".  One way this could be achieved would be to implement a hashmap on top of a tuple in pure python, as bizarre as that sounds.

I took a little stab at this yesterday, also because cuckoo hashing is something I've been interested in for awhile.  However, the implementation is currently buggy and this may be a crazy thing to do in Python regardless:

```python
def cuckoo_insert(x, a):
    pos = hash(x) % len(a)
    if a[pos] is _EMPTY:
        a[pos] = x
        return
    # not sure if this is the best approach?
    pos = pos * (id(a) % len(a)) % len(a)
    if a[pos] is _EMPTY:
        a[pos] = x
        return
    evicted, a[pos] = a[pos], x
    print 'evicted', evicted, 'from', pos
    cuckoo_insert(evicted, a)
    # fingers crossed no infinite loop


def cuckoo_index(x, a):
    pos = hash(x) % len(a)
    if a[pos] == x:
        return pos
    pos = pos * (id(a) % len(a)) % len(a)
    if a[pos] == x:
        return pos
    raise ValueError('{} not in list'.format(x))


def cuckoo_extend(a, b):
    olda = [e for e in a if e is not _EMPTY]
    b = [e for e in b if e is not _EMPTY]
    a[:] = [_EMPTY] * (len(a) + len(b))
    for x in olda + b:
        cuckoo_insert(x, a)


_EMPTY = object()


if __name__ == "__main__":
    init = [_EMPTY] * 100
    for i in range(60) + list('cat'):
        cuckoo_insert(i, init)
    print init```

The way this might work is:

```python
class frozendict(tuple):
   def __init__(self, _dict):
      keys = cuckoo_extend([_EMPTY] * len(_dict) * 3 / 2, _dict)
      values = [_EMPTY] * len(keys)
      for k in _dict:
          values[cuckoo_index(k, keys)] = _dict[k]
      tuple.__init__(self, (tuple(keys), tuple(values)))

   def __getitem__(self, key):
      try:
         ix = cuckoo_index(key, self[0]):
      except ValueError:
         raise KeyError()
      return  self[1][ix]
```

Which is doing `cuckoo_insert()` for every item in the dict!  (Speaking of bad performance right...).

The practical options I can think of:  (1) accept O(n) `__getitem__`; (2) mask off mutation functions from a regular dict and add a `__hash__`.

induane:
Okay so - I've updated the patchset but it's a voracious mess and can probably be made far far more elegant. It does however get into the space that I was thinking about. @mahmoud suggestions for cleanup would be ultra welcome. 

@kurtbrose I ruthlessly stole your idea of subclassing tuple and extended it a bit more with some metaclass programming. The idea is that you get a custom class based on tuple, but with attributes manufactured from calling Type. Thus each frozendict is a custom generated class with attribute values for each of the hash values of it's keys. Each one of those attributes is a lambda that contains a pointer to the index value of the value which it points to.

This has fast pre-indexed lookups and is truly immutable.

boltons/dictutils.py
```diff
@@ -693,4 +693,104 @@ def __reversed__(self):
...
+        while True:
+            try:
+                yield self[SentinelInt(idx)][0]
+            except IndexError:
+                break
+            idx += 1
+
+    def items(self):
+        """Returns a list of tuples of data"""
+        return [(x, self[x]) for x in self]
```
guyarad: also, the implementation here seems to be very inefficient (but since I'm relatively new to Python, I might be missing something). I added this method:

    def alternate_items(self):
        return list(super().__iter__())

And then did that:

    import timeit
    setup = 'from __main__ import frozendict; fd = frozendict({"a": 1, "b": 2, "c": 3})'
    print(min(timeit.repeat('fd.alternate_items()', setup=setup)))  # output: 0.8592938940055319
    print(min(timeit.repeat('fd.items()', setup=setup)))  # output: 10.080946975998813

boltons/dictutils.py
```diff
@@ -693,4 +693,104 @@ def __reversed__(self):
...
+
+    def iteritems(self):
+        """Iterates, returning key, value pairs as a tuple"""
+        idx = 0
+        while True:
+            try:
+                yield self[SentinelInt(idx)][0], self[SentinelInt(idx)][1]
+            except IndexError:
+                break
+            idx += 1
```
guyarad: I'm guessing this method can also be implemented using `super().__iter__()` (but didn't check yet)

induane: This isn't a subclass of a dict so the super calls would be going to the tuple class which won't return the same things.

boltons/dictutils.py
guyarad: I know about the differences between py2 and 3 (as I pointed out). Since this library is suppose to be compatible with both I'm not sure what's the right path.
Your rationale waiting until design/approach is finalize makes perfect sense.

One thing you missed is my comment about the implementation of the items. My suggested implementation also creates a list, that's identical to what `items` produces, but is far more efficient.

boltons/dictutils.py
guyarad: never tried py2.6. Is still being used at all? I guess if this library is intending to support we do have a problem here. Regarding the subclass - I know. Essentially you are creating a tuple of pairs. Iterating of that tuple will return a tuple, that's a key/value pair. That's exactly what's iterating over a dict return (when using `iteritems` and on Py3 `items`).

induane: 2.6 support is necessary for boltons

mahmoud:
So, as mentioned in some of the comments above, this particular approach to a FrozenDict is a bit too tricky for my tastes. I whipped up a version that suits my purposes and included it in boltons==19.0.0: https://boltons.readthedocs.io/en/latest/dictutils.html#boltons.dictutils.FrozenDict

Feel free to take a look at [the implementation](https://github.com/mahmoud/boltons/blob/71307e7db62a80a787caf2f2a6be49810824555b/boltons/dictutils.py#L974-L1035) and let me know if you spot any issues or have any questions. Thanks for your patience and comments, all!

PR #106, opened 2016-11-20
Simplified and optimized iterutils.one.

coady:


mahmoud:
You're a simplifyin' machine! Fewer lines doesn't always mean better, but these changes really do seem like they could stack up for larger inputs. I actually finished my benchmark and code review before appveyor came back, but with all checks passing it's looking good to me!

NB: here's a quick microbenchmark I did to test before/after:
```
$ python -m timeit -s "from boltons.iterutils import one; src = ([False] * 5000) + [True] + ([False] * 500)" "one(src)"
10000 loops, best of 3: 103 usec per loop

$ python -m timeit -s "from new_one import one; src = ([False] * 5000) + [True] + ([False] * 500)" "one(src)"
10000 loops, best of 3: 39.2 usec per loop
```

Thanks again!

PR #107, opened 2016-11-29
Enhance `bucketize` by adding filter_by_key and value_transform

guyarad:
- `filter_by_key`: will allow excluding certain buckets from being collected.
- `value_transform`: will allow doing a transformation on the values before adding them to the bucket.

### DIFF
--- a/boltons/iterutils.py
+++ b/boltons/iterutils.py
@@ -467,10 +467,12 @@ def backoff_iter(start, stop, count=None, factor=2.0, jitter=False):
     return
 
 
-def bucketize(src, key=None):
+def bucketize(src, key=None, value_transform=None, filter_by_key=None):
     """Group values in the *src* iterable by the value returned by *key*,
     which defaults to :class:`bool`, grouping values by
     truthiness.
+    *value_transform* can be used to modify the value before returning it.
+    *filter_by_key*, if defined, will allow excluding certain buckets from being collected.
 
     >>> bucketize(range(5))
     {False: [0], True: [1, 2, 3, 4]}
@@ -483,7 +485,20 @@ def bucketize(src, key=None):
     >>> bucketize([None, None, None, 'hello'])
     {False: [None, None, None], True: ['hello']}
 
-    Note in these examples there were at most two keys, ``True`` and
+    Bucketize into more than 3 groups
+
+    >>> bucketize(range(10), lambda x: x % 3)
+    {0: [0, 3, 6, 9], 1: [1, 4, 7], 2: [2, 5, 8]}
+
+    >>> bucketize(range(5), value_transform=lambda x: x*x)
+    {False: [0], True: [1, 4, 9, 16]}
+
+    Filtering by key
+
+    >>> bucketize(range(10), key=lambda x: x % 3, filter_by_key=lambda k: k % 3 != 1)
+    {0: [0, 3, 6, 9], 2: [2, 5, 8]}
+
+    Note in some of these examples there were at most two keys, ``True`` and
     ``False``, and each key present has a list with at least one
     item. See :func:`partition` for a version specialized for binary
     use cases.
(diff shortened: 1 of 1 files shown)


kurtbrose:
This is an interesting approach.  I wonder though, would it be better to use a list comprehension than add something to the API?

```python
bucketize([value_transform(e) for e in input if filter_by_key(keyfunc(e))], keyfunc)
```
versus

```python
bucketize(input, keyfunc, value_transform, filter_by_key)
```

The first version is denser, but I find it to be a bit more readable just because I don't have to remember as much of `bucketize`'s behavior.  What do you think?  Did you run into a situation where you wanted this functionality?  (It helps me in API design to have a specific use case in mind.)

guyarad:
It basically all comes down to the fact that after value is transformed, the key function isn't valid. So I can't do the transformation outside of the `bucketize` call. 
The `filter_by_key` part is essentially for performance reasons (so I'm less eager to have it added). If I don't need a specific key, why created the relevant list and keep adding to it?

Consider the following scenario:
    
    class Status(Enum):
        created = 1
        in_progress = 2
        completed = 3

    class Task(object):
        def __init__(self, id, status):
            self.status = status
            self.id = id

    def main():
        tasks = ... # get a list[Task] from somewhere
        # I'm interested in bucketize these tasks, with the following properties:
        #   1. Key is the status (which is task.status)
        #   2. The values are only the task IDs (task.id)
        #   3. I don't care about completed tasks (task.status != Status.completed)

        # that's what I need to do without the above enhancement:
        bucketized = bucketize(tasks, key=operator.attrgetter('status'))
        if Status.completed in bucketized:
            del bucketized[Status.completed]
        bucketized = { k : map(operator.attrgetter('id'), ts) for k, v in bucketized.items() }

        # and with it the enhancement:
        bucketized = bucketize(tasks,
                               key=opereator.attrgetter('status'),
                               value_transform=operator.attrgetter('id'),
                               filter_by_key=partial(operator.ne, Status.completed))

kurtbrose:
Thanks for explaining.  It might be possible to code golf it around a bit.

```python
bucketized = { k : [e.id for e in v] for k,v in bucketize(tasks, key=lambda a: a.status) if k != completed}
```

versus

```python
bucketized = bucketize(tasks, key=lambda a: a.status, lambda a: a.id, lambda a: a.status != completed)
```

guyarad:
Feels to me that using the enhanced version is more expressive (I don't need to try and parse out the list comprehension, which is a great tool, but when lines get too long is confusing).

That being said, I did some benchmark, and for some reason my version seems less performant (mine was 28 seconds and original was 26 seconds):
```python
import collections
import functools
import operator
import itertools
import timeit

Task = collections.namedtuple('Task', 'id status')

key_func = operator.attrgetter('status')
key_filter = functools.partial(operator.ne, 2)
val_trans = operator.attrgetter('id')


def make_test_list(elements_count=500):
    statuses = [1, 2, 3]
    return list(itertools.islice(map(Task, itertools.count(), itertools.cycle(statuses)), elements_count))


if __name__ == '__main__':
    tasks = make_test_list()
    old_bucketized = {k: [e.id for e in v] for k, v in old_bucketize(tasks, key=key_func).items() if k != 2}

    bucketized = bucketize(
        tasks,
        key=operator.attrgetter('status'),
        value_transform=operator.attrgetter('id'),
        filter_by_key=functools.partial(operator.ne, 2))

    assert sorted(old_bucketized.keys()) == sorted(bucketized.keys())
    for key in bucketized:
        assert tuple(sorted(bucketized[key])) == tuple(sorted(old_bucketized[key]))

    old_bucketized = '{k: [e.id for e in v] for k, v in old_bucketize(tasks, key=key_func).items() if k != 2}'
    new_bucketized = "bucketize(" \
                     "tasks, " \
                     "key=key_func, " \
                     "value_transform=val_trans, " \
                     "filter_by_key=key_filter)"

    setup = 'from __main__ import bucketize, old_bucketize, make_test_list, key_func, key_filter, val_trans;' \
            'import operator, functools; ' \
            'tasks = make_test_list(50)'

    print(min(timeit.repeat(old_bucketized, setup=setup)))
    print(min(timeit.repeat(new_bucketized, setup=setup)))
```

mahmoud:
Thanks for your patience, @guyarad! While I definitely share Kurt's concern about API creep, I think this is basically a good change. The performance regression you saw was probably due to Python's function call overhead, but at all scales I've used bucketize, it's very unlikely to be an issue. I see you know how to gather timings, so let me know if and how real-world performance becomes a concern. :) thanks again!

guyarad:
Cool! my contribution to a real project :)

mahmoud:
Ah, welcome to the wide world of free software! Also note that I made the argument names more consistent by calling `filter_by_key` into `key_filter` (goes well with `value_transform`) :)

PR #108, opened 2016-12-29
Simplified same.

coady:


mahmoud:
All tests passing, all code simpler. No benchmarks, but at this point I can take it that this is classic high-quality @coady work. Thanks!

Issue #109, opened 2017-01-13
'cachedproperty' decorator replace original's method docstring


kdeldycke:
I'm using ``cachedproperty`` a lot as a decorator but the latter seems to replace the original method's docstring by its own.

See for instance that code: https://github.com/kdeldycke/maildir-deduplicate/blob/v2.0.1/maildir_deduplicate/mail.py#L59-L68 . It is rendered by Sphinx with the decorator's docstring instead of the one from the decorated method. See: https://maildir-deduplicate.readthedocs.io/en/stable/maildir_deduplicate.html#maildir_deduplicate.mail.Mail.message

induane:
You might look at the implementation in Django's functional tools. I
believe its cached property decorator preserves the original docstring.

induane:
I added a pull request that should fix the issue:

https://github.com/mahmoud/boltons/pull/110

mahmoud:
My my my, thank you @kdeldycke for the report and @induane for the fix! I've merged and will release shortly. Thanks again, both of you!

kdeldycke:
Wow, that was fast! Thanks @induane and @mahmoud ! :)

PR #110, opened 2017-01-13
Preserve original docstring in cachedproperty

induane:
This fixes Issue #109

Set the __doc__ attribute on the original
docstring so that the docstring is preserved
and visible to tooling like sphinx and epydoc.

Slightly simply __get__ and remove a line of code


mahmoud:
Great! Tested and ready to roll, just the way I like em. Merged :)

PR #111, opened 2017-01-16
Test boltons against all Python-related Travis CI targets.

kdeldycke:
Changelog:
* Try building boltons against all possible Python targets.
* Document Travis CI Python targets.
* Update list of trove classifiers.
* Document trove classifiers.

This is mainly a test to see if boltons stand up against all Python targets I found on Travis. The reason I started this is I stumble upon an [issue with Python 3.3 and `remap`](https://travis-ci.org/kdeldycke/meta-package-manager/jobs/192211967#L576) utility.

I'll be happy to close or amend this PR depending on feedbacks.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,12 +1,30 @@
 language: python
 sudo: false
 cache: pip
+
+# Python targets, as defined by https://github.com/travis-ci/travis-build/blob
+# /master/spec/build/script/python_spec.rb and https://github.com/travis-ci
+# /travis-build/blob/master/lib/travis/build/script/python.rb
 python:
-  - "2.6"
+  # Standard release https://docs.travis-ci.com/user/languages
+  # /python#choosing-python-versions-to-test-against
   - "2.7"
+  - "3.3"
   - "3.4"
   - "3.5"
-  - "pypy"
+  - "3.6"
+  - "3.7-dev"  # a.k.a "nightly"
+
+  # PyPy2.7: https://doc.pypy.org/en/latest
+  # /index-of-release-notes.html#cpython-2-7-compatible-versions
+  - pypy  # PyPy2.7 2.5.0
+  - pypy-5.1
+  - pypy-5.3
+  - pypy-5.4
+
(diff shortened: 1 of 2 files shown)


kdeldycke:
So as I suspected, boltons works on the whole range of Python targets but Python 3.3. Should we fix the issue? FYI, [Python 3.3 is supported until 2017-09](https://en.wikipedia.org/wiki/CPython#Version_history).

induane:
I put up a patch to fix the Python3.3 compatibility problem: https://github.com/mahmoud/boltons/pull/116

kdeldycke:
Thanks @induane for the PR! :)

mahmoud:
Thanks to both of you for the great work so far. This is a good thing in principle, and I'm not against supporting it, but we do need to get the tests passing. I can probably look at this tomorrow, but in case someone wants to beat me to it, I think we're being held up by a single doctest: https://travis-ci.org/mahmoud/boltons/jobs/192215564#L232

Also, just to be up front, I am *for* removal of Python 3.3 support much sooner than, e.g., 2.6. Adoption and usage make 3 a totally different beast than 2. I can comment further if more explanation is needed ;)

kdeldycke:
Thanks @mahmoud for the merge!

No need to argue for Python 3.3. It's going to be unmaintained in less than 8 months anyway. If I'm supporting it in my packages it's because of the low-overhead it incurs.

I don'y see any issue removing Python 3.3 support in Boltons'. I'll simply drop it too in my packages.

kdeldycke:
Any plan to cut a release soon? 😇

mahmoud:
@kdeldycke ah yes, I was trying to get a feature done in time, but in the end it's best not to rush. 17.0.0 is fully released (you might want to update that requirement to reflect this). [Changelog here](https://github.com/mahmoud/boltons/blob/master/CHANGELOG.md#1700)! Thanks again everyone!

kdeldycke:
Fantastic! Thanks for the quick release! 😃

Issue #112, opened 2017-01-16
tbutils ExceptionInfo and TracebackInfo to_dict() result not JSON serializable


DoronCohen:
I wanted to use tbutils to serialize whatever exceptions thrown at me and store them in a DB but `to_dict()` result is not JSON serializable which is kind of weird.

When I looked it up it seems that it's because `line` field in each `CallPoint` object is not of a native `str` type (although it says it is in the [docs](http://boltons.readthedocs.io/en/latest/tbutils.html#boltons.tbutils.Callpoint)). It is of type `boltons.tbutils._DeferredLine`.

I wrote a JSON serializer that handles that, but if there is no special reason for that I suggest returning a native string. Or at least update the documentation.

induane:
I can see two potential solutions to this:

The first is to make the resulting dictionary contain string instances of _DeferredLine. After all, the docstring for ``to_dict`` says "Get a :class:`dict` copy of the Callpoint. Useful for serialization."

This would effectively change the API though; I'm not sure what @mahmoud  feels in regards to API stability here. There is always a slim chance this could break things for someone who is using to_dict in an unconventional or unexpected way.

This leaves another option:  Provide a ``to_json`` method for all Callpoint object instances. This would result in a stable API and JSON serialization support. Potentially handy.

I actually lean towards the first solution personally because it seems weird to have a class designed for serialization that can't actually be easily serialized. 

To that end I've created two pull requests, each implementing one version of the fix.

https://github.com/mahmoud/boltons/pull/113

and

https://github.com/mahmoud/boltons/pull/114

Really I'm not in a position to make the call here, but I figured it was easy enough to provide both potential solutions and let the boltons maintainers make a call.

mahmoud:
Yeah, I'm probably the most extensive user of this API, but having gone through all my uses, they all stringify the DeferredLine, so not only won't the first solution break my code, but it probably should have been built-in to begin with. I'm going to do some more testing before release, but for now I think I'm going to go with #114.

mahmoud:
#114 is merged and a release is forthcoming. Thank you both for the report and the great fixes!

PR #113, opened 2017-01-16
Add to_json method to Callpoint class

induane:
This is one possible solution to the issue outlined in https://github.com/mahmoud/boltons/issues/112
I have another pull request with a different solution.

Added a to_json method to the Callpoint
class for serialization.


mahmoud:
Closing in favor of #114, thank you for both!

PR #114, opened 2017-01-16
Make Callpoint.to_dict result JSON serializable

induane:
This is one possible solution to the issue outlined in https://github.com/mahmoud/boltons/issues/112
I have another pull request with a different solution.

Converts all instances of _DeferredLine into string
objects when returning the dictionary. This has the
effect of making the resulting dictionary instance
JSON serializable.


mahmoud:
Looks good! Merging to fix #112.

PR #119, opened 2017-02-13
Change Sentinel __nonzero__ to __len__ for Python3 Compatibility

zeroSteiner:
This fixes a Python3 compatibility bug in the `make_sentinel` function. Python3 uses `__bool__` instead of Python2's `__nonzero__`. According to the docs<sup>[1]</sup> they will both fail back to `__len__`, checking for a non-zero value. This simple change makes the `Sentinel` always evaluate to False in both Python2 and Python3.

Verification steps:
- [ ] Create a sentinel in Python2 and verify that it evaluates to False
- [ ] Create a sentinel in Python3 and verify that it also evaluates to False

Example output using Python 3.5.2
```
>>> sys.version_info
sys.version_info(major=3, minor=5, micro=2, releaselevel='final', serial=0)
>>> import boltons.typeutils
>>> disabled = boltons.typeutils.make_sentinel('DISABLED')
>>> True if disabled else False
False
```

Thanks!

<sup>[1]</sup> See [Python2 \_\_nonzero\_\_](https://docs.python.org/2/reference/datamodel.html#object.__nonzero__) and [Python3 \_\_bool\_\_](https://docs.python.org/3/reference/datamodel.html#object.__bool__)

### DIFF
--- a/boltons/typeutils.py
+++ b/boltons/typeutils.py
@@ -53,8 +53,8 @@ def make_sentinel(name='_MISSING', var_name=None):
             def __reduce__(self):
                 return self.var_name
 
-        def __nonzero__(self):
-            return False
+        def __len__(self):
+            return 0
     return Sentinel()
 
 


induane:
Alternatively you could simply implement bool

since __nonzero__ is already implemented you could just add

__bool__ = __nonzero__

The behavior is going to work out the same, but the code is more obvious to read; by implementing the boolean value using __len__ you rely on it being obvious that __len__ is the fallback to bool.

SzieberthAdam:
@induane: Except that it pollutes a Python 3 object with a magic method which is removed from that version. I would recommend an `if PY_VER` block instead.

induane:
@SzieberthAdam `__bool__ = __nonzero__` is effectively just an alias; doesn't seem too polluted to me as I tend to subscribe some to the sort of "Explicit is better than implicit." concept. In the end it doesn't matter too much, it's just a style choice and everyone's got their own sense of aesthetics.

SzieberthAdam:
True, and I respect your opinion. In addition, this is minor difference.

mahmoud:
Hi all! I'm glad all the bikeshedding settled out before I had a chance to check this out. Too busy adding an as-yet-undocumented [urlutils module](https://github.com/mahmoud/boltons/blob/master/boltons/urlutils.py). Check it out if you're inclined!

As for the great debate, I merged the PR as an acknowledgement of the issue, but switched the implementation to @induane's approach. I might have gone with @SzieberthAdam if there'd already been a `_PY3` or something in the file, but really dunder pollution on miscellaneous objects isn't something we can really complain about in Python :)

Oh, and I added a test, too. Anyways, thank you all for your contributions!

PR #120, opened 2017-02-15
Bring cachedproperty fixes to urlutils

induane:
I noticed that the cachedproperty decorator was in the new urlutils module but didn't have the previously applied fixes. This patch brings those changes in.


mahmoud:
Yeah, that's a good idea. I probably would've forgotten to update this before release otherwise. Thanks!

PR #128, opened 2017-06-18
fix the default arguments of mathutils.clamp

tiwo:
`mathutils.clamp` is supposed to restrict an arbitrary number to the interval given by its last two (optional) arguments. But it is incorrect when not both `lower` and `upper` arguments are given

The default arguments are `None`, which isn't comparable in Python 3 and is less than x for any x in Python 2 (except NaN).

| expression  | Python 2 | Python 3 | fixed |
|---|---|---|---|
| `clamp(1)` |  `None` | raises `TypeError`  | `1` |
| `clamp(1, upper=2)` | `1`  | raises `TypeError` | `1` |
| `clamp(1, lower=0)` | raises `ValueError`  | raises `TypeError` | `1` |

This fixes `mathutils.clamp` by simply using positive infinity and negative infinity as default bounds, and adds two test cases as well as an example in the function's docstring.

### DIFF
--- a/boltons/mathutils.py
+++ b/boltons/mathutils.py
@@ -23,6 +23,8 @@ def clamp(x, lower=float('-inf'), upper=float('inf')):
     0
     >>> clamp(101.0, 0, 5)
     5
+    >>> clamp(123, upper=5)
+    5
 
     Similar to `numpy's clip`_ function.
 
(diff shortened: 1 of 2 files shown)


mahmoud:
Excellent points and excellent contribution! Very nice work, @tiwo, thank you!

PR #129, opened 2017-06-21
Support passing dir argument to TemporaryFile rollover

sp1ky:


mahmoud:
Hey David! Just took a look and this seems pretty good and useful so far, thanks! I think it'd be good if the `dir` argument was `tmpdir` or `tmp_path` or something a bit more descriptive. I'm also going to ping the original Spooled* author, @induane, in case he has any comments. Thanks again!

tests/test_ioutils.py
```diff
@@ -40,6 +41,29 @@ def test_getvalue_rollover(self):
         self.assertEqual(self.spooled_flo.getvalue(), self.test_str)
         self.assertTrue(self.spooled_flo._rolled)
 
+    def test_rollover_custom_directory(self):
+        """dir keyword argument is passed to TemporaryFile instantiation"""
+        custom_dir = mkdtemp()
+        try:
```
harvimt: can be `with tempfile.TemporaryDirectory() as custom_dir:` ... instead of `mkdtemp` + `try...finally` (saves like 3 lines of code!)

https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryDirectory

sp1ky:
Thanks for the reviews. 

@mahmoud Sure thing. I was mirroring the TemporaryFile keyword argument 'dir' but I guess there's no real reason to stick to it. I'll update the pull request with something like "temp_dir".

@harvimt Unfortunately TemporaryDirectory() only exists in Python 3.2+. Thanks for the suggestion though!

mahmoud:
@sp1ky ah, I missed that it was a symmetry/consistency thing. That's the sort of compromise I can appreciate, so I'm fine with a `dir` kwarg. Sorry to flip flop, but feel free to switch it back! I might do so myself if you haven't gotten to it by the time I'm home tonight (~9 hours from time of writing)

mahmoud:
All merged! Thanks for the code and review, everyone!

sp1ky:
Great! Thanks

PR #130, opened 2017-06-23
Update ioutils documentation

sp1ky:
* Fix some typos
* Remove duplicated code from the zipfile download example
* Update the zipfile download example to use requests context manager


mahmoud:
Excellent! Thank you!

PR #132, opened 2017-07-02
Small docstring typo correction for OMD.

yassineAlouini:
In this PR, I correct a minor typo from the `OrderedMultiDict` docstring.


mahmoud:
Good catch, thank you!

Issue #133, opened 2017-07-07
`boltons.funcutils.wraps` breaks with Python 3 function annotations


rowillia:
Tested with `boltons==17.1.0`

Repro: Executing the following script fails
```python
from boltons.funcutils import wraps
from typing import Optional

def wrap_it(f):
    @wraps(f)
    def wrapper(self, *args, **kwargs):
        f(*args, **kwargs)
    return wrapper

@wrap_it
def foo(x: Optional[int]) -> Optional[int]:
    return x
```

Results in
```bash
rwilliams at rwilliams-mbp114 in ~/src/oss/boltons (master●) (venv)
$ python ../../test_wraps_with_annotations.py
Traceback (most recent call last):
  File "../../test_wraps_with_annotations.py", line 11, in <module>
    def foo(x: Optional[int]) -> Optional[int]:
  File "../../test_wraps_with_annotations.py", line 5, in wrap_it
    @wraps(f)
  File "/Users/rwilliams/src/oss/boltons/boltons/funcutils.py", line 307, in wrapper_wrapper
    fully_wrapped = fb.get_func(execdict, with_dict=update_dict)
  File "/Users/rwilliams/src/oss/boltons/boltons/funcutils.py", line 526, in get_func
    self._compile(src, execdict)
  File "/Users/rwilliams/src/oss/boltons/boltons/funcutils.py", line 590, in _compile
    exec(code, execdict)
  File "<boltons.funcutils.FunctionBuilder-0>", line 1, in <module>
NameError: name 'Union' is not defined
```

mahmoud:
Ah, using the fancy stuff! Good to see! Historically @markrwilliams oversees the Python 3 version of this, and I think it's supposed to support these features. Any quick ideas Mark?

rowillia:
@mahmoud @markrwilliams This is what I came up with: https://github.com/mahmoud/boltons/pull/134 .  I know it's kind of gross, but it respects the letter of PEP 484 and I'm not familiar enough with boltons to know how to properly solve it.

episodeyang:
Looks like I forgot to check the open issues! #138 

I'm running into similar issue.

```error
Traceback (most recent call last):
  File "/Users/ge/machine_learning/meta_learning_project/MAML_tensorflow/maml_bradly.py", line 273, in <module>
    def demo_fn(params: Some):
  File "/Users/ge/anaconda/envs/simulation/lib/python3.6/site-packages/moleskin/moleskin.py", line 31, in timeit
[...]

PR #134, opened 2017-07-07
Allow symbols to be used in type annotations

rowillia:
PEP 484 states that string literals can be used in place of type declarations:
https://www.python.org/dev/peps/pep-0484/#the-problem-of-forward-declarations

While this is kind of gross, there aren't really any better options unless we want
to blit in `sys.modules` and even then getting that right would be extremely tricky
if we also consider aliases.

### DIFF
--- a/boltons/funcutils.py
+++ b/boltons/funcutils.py
@@ -401,8 +401,16 @@ class FunctionBuilder(object):
         @classmethod
         def _argspec_to_dict(cls, f):
             argspec = inspect.getfullargspec(f)
-            return dict((attr, getattr(argspec, attr))
-                        for attr in cls._argspec_defaults)
+            result = dict((attr, getattr(argspec, attr))
+                          for attr in cls._argspec_defaults)
+            escaped_annotations = {}
+            for annotation, value in argspec.annotations.items():
+                if isinstance(value, type):
+                    escaped_annotations[annotation] = value.__name__
+                else:
+                    escaped_annotations[annotation] = str(value)
+            result['annotations'] = escaped_annotations
+            return result
 
     _defaults = {'doc': str,
                  'dict': dict,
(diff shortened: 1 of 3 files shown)


mahmoud:
I think it's a valuable approach! We'll have to figure out something for the 2.6 CI, but personally I'm ok with the string usage. Let's see what @markrwilliams thinks.

markrwilliams:
Checking in :) I'm going to push a commit that introduces [environment markers](https://www.python.org/dev/peps/pep-0508/#environment-markers) to `requirements.txt`.  This should work with pip >= 6.0, which [Travis uses for our 2.6 builds](https://travis-ci.org/mahmoud/boltons/jobs/250995226#L173)

@rowillia Have you tested this with mypy?  [Forward references must "evaluate without errors once the module is loaded"](https://www.python.org/dev/peps/pep-0484/#forward-references).  I wonder what happens, then, if we have `module_a.py`:

```python
import typing

def func(blah -> typing.List[int]) -> List[int]:
    pass
```

and we do this in `module_b.py`:

```python
from module_a import func as original_func

@wraps(f)
def func(*args, **kwargs):
    pass
```

mahmoud:
Haha, and I don't know nearly enough about the annotation system. You're definitely my local expert, @rowillia :)

I can say that the eval happens because there's not a good way (if any) to programmatically rewrite a Python function signature. That's sort of the reason behind FunctionBuilder. If annotations are like `__doc__` then, yeah, they can be stripped off and readded. No need to include them at eval time. If you'd like to try that tack, let me know. Otherwise, Mark's got the CI setup working again, so I'm happy to merge!

rowillia:
@mahmoud let me try stashing the annotations first.

mahmoud:
@rowillia standing by. :)

episodeyang:
@mahmoud @markrwilliams @rowillia  Is this ready to be merged?

i'm running into the same issue #133 #138 :joy:

mahmoud:
As of 19.0.1, this should be fixed! See #203 for deets, but thanks to all (esp @itamarst)!

PR #135, opened 2017-07-27
ioutils: add MultiFileReader to ease concatenation of multiple readers

durin42:
I ended up needing something like this for Mercurial, and mhashemirc
suggested that it would make sense in boltons.

### DIFF
--- a/tests/test_urlutils.py
+++ b/tests/test_urlutils.py
@@ -313,6 +313,13 @@ def test_navigate():
     assert navd.to_text() == _dest_text
 
 
+# TODO: RFC3986 6.2.3 (not just for query add, either)
+# def test_add_query():
+#     url = URL('http://www.example.com')
+#     url.qp['key'] = 'value'
+#     assert url.to_text() == 'http://www.example.com/?key=value'
+
+
 def test_self_normalize():
     url = URL('http://hatnote.com/a/../../b?k=v#hashtags')
     url.normalize()
--- a/TODO.rst
+++ b/TODO.rst
@@ -1,15 +1,6 @@
 TODO
 ====
 
-- dummy context manager
-- dummy file
-- urlutils
-
-cacheutils
-----------
-
-- counting dictionary (with sys.modules example)
-
 dictutils
 ---------
 
@@ -32,8 +23,6 @@ misc?
   everything callable, with a hook. Needs an enable/disable flag.
   - get/set/call/return/exception
   - __slots__
-
-- Tracking proxy. An object that always succeeds for all operations, saving the call history.
 - Top/Bottom singletons (greater than and less than everything)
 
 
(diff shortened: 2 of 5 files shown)


mahmoud:
Looks great, @durin42, thanks! I'll handle the 2.6 compat, it's nbd. I'll also take a swing at getting docs in there, too. I'll @ you if there's questions.

mahmoud:
So, @durin42, while writing some docs on this, it occurred to me that we only take `BytesIO` and `TextIOBase`, and for Python 2 we should probably take `file` objects (since that's the class name and `__init__` arg name, after all). Does it sound ok to wrap up passed-in `file` objects with a `BytesIO`?

induane:
@durin42 this looks pretty awesome btw!

@mahmoud  It might make sense to accept SpookedBytesIO and SpooledStringIO instances in the multifile reader as well. It might be confusing to disallow them given that they're file-like-objects in the same module.

durin42:
@mahmoud Hmm, file objects can be open in either bytes or unicode mode in Python 3. Sniff the mode attribute and look for a 'b' to verify it's bytes mode, otherwise it's reading unicodes.

It should be fine to not have to wrap the files at all - just verify they'll return the right kind of data and then leave them in the right place in the sequence (assuming they support seek).

Also agree that handling the Spooled{Bytes,String}IO instances seems reasonable. I didn't think of that because I didn't have such a requirement in hg. :)

tests/test_urlutils.py
```diff
@@ -313,6 +313,13 @@ def test_navigate():
     assert navd.to_text() == _dest_text
 
 
+# TODO: RFC3986 6.2.3 (not just for query add, either)
```
durin42: Wrong branch?

mahmoud: heh, galdangit.

mahmoud:
@durin42 On Python3, `open` returns `io` objects:

```
>>> isinstance(open('test.md'), io.TextIOBase)
True
```

Interestingly, even with 'rb', we get a `TextIOBase`:

```
>>> isinstance(open('test.md', 'rb'), io.TextIOBase)
True
>>> isinstance(open('test.md', 'rb'), io.BytesIO)
False
```

So, depending on your definition, it should "just work" for Python 3. It's mostly handling `file` on py2, which always returns bytes.

durin42:
Ugh, feels like a Python3 bug that a binary-mode file isinstance TextIOBase.

On Python 2, yeah, this is already solved. On Python 3, I guess we have to look for a "mode" attr and then sniff it to figure out if it's a bytes-like or text-like file.

:(

mahmoud:
@durin42 gotcha, all fixed. i think this is basically ready to merge. Added some docs, too. A quick glance at those would be appreciated, too.

mahmoud:
@induane the current version also supports `Spooled*IO`, but I should note that not taking an `initial_value` in the constructor, and having the string first argument silently becoming `max_size` threw me for a bit of a loop. It'd be nice to get a PR to address that, but we can discuss that in a separate issue! :)

durin42:
Docs look good to me. Thanks for finishing off my prototype!

PR #137, opened 2017-08-19
added one-to-one class to dictutils

kurtbrose:


kurtbrose:
oops wait there is a bug with update() -- it doesn't preserve the invariant that each value can have only one key

mahmoud:
Seems pretty useful! The failure is due to a repr change in 3.7 nightly i've fixed on master, so no sweat there. Merging, thank you!

PR #139, opened 2017-09-05
DeferredValue: fix caching

immerrr:
It appears as if DeferredValue has been broken since the very beginning.

It tried caching the value by default, but never accessed it raising a NameError when `get_value` was invoked for the second time.


mahmoud:
Interesting! It's certainly been working [for my purposes](https://github.com/mahmoud/lithoxyl/blob/644083668b32d57692b5de026d8643a1797c66ce/lithoxyl/tests/test_formatters.py#L52), but this definitely looks like an improvement, thank you!

PR #143, opened 2017-10-30
Add threadutils with a TwoWayEvent object

zeroSteiner:
This adds a `threadutils` module with (currently) one new class. The new class, `TwoWayEvent` is a take on Python's builtin `threading.Event` class but allows for threads to wait on the event's internal flag to be either true or false. This is accomplished through the addition of the `wait_clear` method which does the opposite of the standard `threading.Event.wait` method.

For testing:
- [ ] Initialize a new instance of the class
- [ ] Calling `wait(timeout=3)` should pauses while calling `wait_clear(timeout=3)` returns immediately
- [ ] After calling `set()` on the new instance, the inverse of step 2 is present

### DIFF
--- a/boltons/threadutils.py
+++ b/boltons/threadutils.py
@@ -1,0 +1,43 @@
+"""This module provides useful functions on top of Python's
+built-in :mod:`threading` module.
+"""
+import threading
+
+
+class TwoWayEvent(threading.Event):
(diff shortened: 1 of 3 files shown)


zeroSteiner:
Closing this because it would break compatibility with older versions of Python where `threading.Event` is not a class.

PR #144, opened 2017-11-08
Fix line-by-line file iteration in ioutils modules

induane:
- iterator on self was yielding only the first newline
- added __next__ to match next() for py3 compat
- updated tests


mahmoud:
LGTM! Thanks for another great PR :)

PR #150, opened 2018-02-04
Add subdict function

acatton:
This adds an utility function to compute a "sub-dictionnaries" to `dictutils.py`.

I've worked on a project with @mdomke and he was adding a very similar function to all of his projects utility module. At first I thought "whatever dude..." But then I realized that it actually is a very useful function, and I ended up adding the same function to all my projects `utils.py`.

I think this would be a very useful addition to boltons

Disclaimer: I wrote this function from scratch, just copying the function signature/API from @mdomke's original function. I made sure to use a different python style than him, so that no copyright infringement could be claimed. (Even though, I think @mdomke would not mind if I would've just had copy/pasted his original code.)

### DIFF
--- a/boltons/dictutils.py
+++ b/boltons/dictutils.py
@@ -804,6 +804,8 @@ def subdict(d, keep=None, drop=None):
     is a 'sub-dictionnary' of _B_, that means that all keys of _A_ are present
     in _B_.
 
+    This does NOT mutate the original dictionnary in any way.
+
     >>> from pprint import pprint as pp
     >>> pp(subdict({'a': 1, 'b': 2}))
     {'a': 1, 'b': 2}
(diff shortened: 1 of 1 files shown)


mdomke:
You're right @acatton, I would be cool with you reusing the original code… I'm glad that you came up with the idea adding this function to boltons.

mahmoud:
I agree with you both, this looks good, thanks! Would you be ok if I renamed `restrict_to`/`omit` to `include`/`exclude`? I think those names might be more intuitive, but I'm open to others, too.

mdomke:
I think semantically it is okay to replace `omit` with `exclude`. On the other hand I would say that `restrict_to` and `include` are not the same. When having set theory in mind `restrict_to` translates for me as "take the subset with these keys", whereas `include` would intuitively translate into some kind of union operation (even though it doesn't make sense from a functionality point of view). But that's just my opinion…

mahmoud:
Hmm, true. How about `keep` and `skip` (or `drop`)?

acatton:
I personally like `keep` and `drop`. I updated the code.

mahmoud:
Freshly released in 18.0.0, thank you again for your contribution!

PR #152, opened 2018-03-03
Correctly handle RFC2732 IPv6 address literals with ports

zeroSteiner:
It does not appear that the `boltons.urlutils.URL` class is properly handling IPv6 literal addresses (as defined in [RFC 2732](https://tools.ietf.org/html/rfc2732)) that specify ports. This is due to how the host string is being parsed in `boltons.urlutils.parse_url`.

For example before this patch (example specifies IPv6 address `::1` with port `8080`):
```
In [2]: pprint.pprint(urlutils.parse_url('tcp://[::1]:8080/'))
{'_netloc_sep': '//',
 'authority': '[::1]:8080',
 'family': None,
 'fragment': None,
 'host': '[::1]:8080',
 'password': None,
 'path': '/',
 'port': None,
 'query': None,
 'scheme': 'tcp',
 'username': None}
```

After this patch:
```
In [1]: import pprint; from boltons import urlutils

In [2]: pprint.pprint(urlutils.parse_url('tcp://[::1]:8080/'))
{'_netloc_sep': '//',
 'authority': '[::1]:8080',
 'family': <AddressFamily.AF_INET6: 10>,
 'fragment': None,
 'host': '::1',
 'password': None,
 'path': '/',
 'port': 8080,
 'query': None,
 'scheme': 'tcp',
 'username': None}
```

Note the differences in the `family`, `host`, and `port` keys of the resulting dictionary.

### DIFF
--- a/boltons/urlutils.py
+++ b/boltons/urlutils.py
@@ -892,16 +892,19 @@ def parse_url(url_text):
     if hostinfo:
         host, sep, port_str = hostinfo.partition(u':')
         if sep:
-            if u']' in port_str:
-                host = hostinfo  # wrong split, was an ipv6
-            else:
-                try:
-                    port = int(port_str)
-                except ValueError:
-                    if port_str:  # empty ports ok according to RFC 3986 6.2.3
-                        raise URLParseError('expected integer for port, not %r'
-                                            % port_str)
-                    port = None
+            if host[0] == u'[' and u']' in port_str:
+                host_right, _, port_str = port_str.partition(u']')
+                host = host + u':' + host_right + u']'
+                if port_str and port_str[0] == u':':
+                    port_str = port_str[1:]
+
+            try:
+                port = int(port_str)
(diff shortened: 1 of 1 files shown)


mahmoud:
Hey Spencer, thanks for this contribution! Lately I've been working a lot on URLs, but mostly [over here in the hyperlink repo](https://github.com/python-hyper/hyperlink). Hyperlink's a merger between urlutils and twisted's URL handling code. I've been meaning to backport many of the fixes, but haven't had the time yet!

I'll definitely review this before the next boltons release, but I just wanted to drop you a note that it may be a better idea to use hyperlink for now. It's not a drop-in API, but it's much better tested and in a better place to be improved, too :)

mahmoud:
Been long minute, but as this PR doesn't seem to hurt anything, and looks pretty good, consider it merged! Thank you for your work and your patience!

PR #154, opened 2018-04-24
Cache utils yield keyerrors when at limit.

induane:
- The LRU and LRI caches hit a key error when trying
  to evict; this patch adds a test for this behavior.

I haven't dug into a fix for this yet.

### DIFF
--- a/tests/test_cacheutils.py
+++ b/tests/test_cacheutils.py
@@ -30,6 +30,20 @@ def test_lri():
     assert len(bc) == 10
 
 
+def test_lri_eviction():
+    cache = LRI(max_size=10)
+    for char in string.ascii_letters:
+        cache[char] == char.upper()
+    assert len(cache) == 10
+
+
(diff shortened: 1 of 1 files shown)


induane:
I've opened an issue report for this at #155

acatton:
@induane it looks like your fix broke the tests, right?

induane:
@acatton I hadn't actually fixed it - turns out the test case I wrote had a typo though; the test case I wrote actually passes with the typo removed.

There is a cache eviction problem though, it just turns out it's only with the LRI cache. My colleague @sp1ky is prepping a full fix for the issue now.

sp1ky:
#157

mahmoud:
Wow, I have been chuggin away with PyCon prep, but you guys handled this wonderfully! I'm going to merge in #157 and close out everything else, let me know if I did something wrong. Thanks for the excellent work, everyone!

Issue #155, opened 2018-04-24
LRU and LRI yield key errors when max_size reached.


induane:
I've created a testcase for the behavior at pull request #154 

When the max size is reached, attempting to set a new key yields a KeyError. I'm not sure why exactly - the error comes from calling ``del self[old]`` - somewhere up the chain this triggers the problem.

sp1ky:
#157

mahmoud:
Thanks for the report! Gonna discuss this over in #157. :)

PR #157, opened 2018-05-04
Fix issue evicting duplicate keys from LRI (#155)

david-bannon:
The insertion deque() can have duplicate items when the same key is set
multiple times. The eviction routine assumes the cache item still exists
and so can raise KeyError. This patch modifies the LRI setitem method
to use dict.pop(key) instead of del[key] so as to avoid the KeyError.

The eviction is now run in a while loop until the cache is less than or
equal to max_size since the dict.pop() eviction is not guaranteed to
actually evict anything.

### DIFF
--- a/boltons/cacheutils.py
+++ b/boltons/cacheutils.py
@@ -351,10 +351,11 @@ class LRI(dict):
 
     def __setitem__(self, key, value):
         # TODO: pop support (see above)
-        if len(self) >= self.max_size:
-            old = self._queue.popleft()
-            del self[old]
         super(LRI, self).__setitem__(key, value)
+        while len(self) > self.max_size:
+            # Keep attempting to remove the least recently inserted
+            # items until we are under max_size.
+            self.pop(self._queue.popleft())
         self._queue.append(key)
 
     def update(self, E, **F):
--- a/tests/test_cacheutils.py
+++ b/tests/test_cacheutils.py
@@ -97,6 +97,13 @@ def test_lru_basic():
     assert second_lru != lru
 
 
+def test_lri_with_dupes():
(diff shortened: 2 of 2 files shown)


mahmoud:
The reason here is hinted at in the comment about `pop()`, but I guess we never got around to fixing it. `__delitem__` and `pop` don't have an efficient way to scan the deque to remove the cached value, so it just... doesn't. For the case of duplicate values, as tested here, a similar issue is happening, where we don't know if the same value was reinserted later. A dict of counts might fix this. I'll have to think about it a bit more :)

crazymerlyn:
> For the case of duplicate values, as tested here, a similar issue is happening, where we don't know if the same value was reinserted later.

How about not duplicating the keys in the first place?
I think replacing the code at [__setitem__](https://github.com/mahmoud/boltons/blob/874e930bd6dbd92e64412e1bdb46a4053587c86d/boltons/cacheutils.py#L357) with
```python
if key not in self:
    self._queue.append(key)
super(LRI, self).__setitem__(key, value)
```
should work.

CameronCairns:
The issue with that solution is you introduce a linear time search on the queue whenever you assign a value to a key (`key not in self` requires traversing the queue).  The original implementation only introduced a constant time overhead to the key value assignment process.

Personally I'm leaning more towards a solution that mimics the linked list approach used in the LRU. It has a little bit of a higher overhead memory footprint, but assignment and cache eviction would take a constant amount of time.

mahmoud:
It's taken a while to arrive at the correct course of action, but as @CameronCairns described, this issue has finally been addressed in 19.0.0, just released! Thanks to you all for your work, and your patience :)

Issue #158, opened 2018-05-31
URL.navigate fails with path in base URL


wcooley:
The following and several permutations fail attempting to concatenate a `tuple` with a `list`:

```
URL('http://localhost').navigate('a').navigate('b')
```
Result:

```
...
            else:  # relative path
>               new_path_parts = list(self.path_parts[:-1]) + dest.path_parts
E               TypeError: can only concatenate list (not "tuple") to list

boltons/urlutils.py:657: TypeError
```

I am a little unclear about exactly what *should* happen when 'a' has a trailing slash or not, but enough cases fail that I know *something* is wrong.
I will follow shortly with a PR with a number of test cases, some of which I know are wrong, so I am hoping I can get some clarification about what the correct behavior *should* be.

This is not strictly an issue with chaining the `navigate` calls, but rather ambiguity of what should happen with the base URL lacks a trailing '/' and failures that result from that.

Env data:
platform darwin
Python 3.6.5 (built w/pyenv)
pytest-3.5.1

PR #160, opened 2018-05-31
Fix TypeError in issue #158/#159

wcooley:
This makes only one more test pass (now only 3 failing) but at least prevents the `TypeError` when attempting to add a `list` and a `tuple`.

pytest output attached
[pytest-out-2.txt](https://github.com/mahmoud/boltons/files/2060262/pytest-out-2.txt)

### DIFF
--- a/tests/test_urlutils.py
+++ b/tests/test_urlutils.py
@@ -268,6 +268,23 @@ def test_rel_navigate():
     return
 
 
+@pytest.mark.parametrize(
+    ('expected',            'base',             'a',    'b'), [
+    ('https://host/b',      'https://host',     'a',    '/b'),
+    ('https://host/a/b',    'https://host',     'a',    'b'),
+    ('https://host/a/b',    'https://host',     'a/',   'b'),
+    ('https://host/a/b',    'https://host',     '/a',   'b'),
+    ('https://host/a/b',    'https://host/a/',  None,   'b'),
+    ('https://host/a/b',    'https://host/a',   None,   'b'),
+])
+def test_chained_navigate(expected, base, a, b):
+    """Chained :meth:`navigate` calls produces correct results."""
+    if a is not None:
+        assert expected == URL(base).navigate(a).navigate(b).to_text()
+    else:
+        assert expected == URL(base).navigate(b).to_text()
+
+
 def test_navigate():
     orig_text = u'http://a.b/c/d?e#f'
     orig = URL(orig_text)
--- a/boltons/urlutils.py
+++ b/boltons/urlutils.py
@@ -654,7 +654,8 @@ class URL(object):
             if dest.path.startswith(u'/'):   # absolute path
                 new_path_parts = list(dest.path_parts)
             else:  # relative path
-                new_path_parts = self.path_parts[:-1] + dest.path_parts
+                new_path_parts = list(self.path_parts[:-1]) \
+                               + list(dest.path_parts)
         else:
             new_path_parts = list(self.path_parts)
             if not query_params:


mahmoud:
It's taken a while to arrive at the correct course of action, but this pickle of an issue has finally been fixed in 19.0.0, just released. Thanks for your work, and your patience :)

mahmoud:
Thought I was closing a different issue, reopening!

tests/test_urlutils.py
```diff
@@ -268,6 +268,23 @@ def test_rel_navigate():
     return
 
 
+@pytest.mark.parametrize(
+    ('expected',            'base',             'a',    'b'), [
+    ('https://host/b',      'https://host',     'a',    '/b'),
+    ('https://host/a/b',    'https://host',     'a',    'b'),
```
allanlei, line 274: Using
```
boltons.urlutils.URL(boltons.urlutils.URL('https://host').navigate('a')).navigate('b')
```
the expected result should be `https://host/b`

tests/test_urlutils.py
```diff
@@ -268,6 +268,23 @@ def test_rel_navigate():
...
+    ('https://host/a/b',    'https://host',     'a/',   'b'),
+    ('https://host/a/b',    'https://host',     '/a',   'b'),
```
allanlei, line 276: Using
```
boltons.urlutils.URL(boltons.urlutils.URL('https://host').navigate('/a')).navigate('b')
```
the expected result should be `https://host/b`

tests/test_urlutils.py
```diff
@@ -268,6 +268,23 @@ def test_rel_navigate():
...
+    ('https://host/a/b',    'https://host/a/',  None,   'b'),
+    ('https://host/a/b',    'https://host/a',   None,   'b'),
```
allanlei, line 278: Using
```
boltons.urlutils.URL(boltons.urlutils.URL('https://host').navigate('/a')).navigate('b')
```
the expected result should be `https://host/b`

allanlei:
@mahmoud @wcooley I've extended this PR in #298.

tests/test_urlutils.py
wcooley: I haven't looked at this in a long time, so pardon me if I'm completely backwards and no longer clear on what I thought behavior should be.

When you say, "the expected result should be", where are you basing that "should be" on?  The scant documentation says, "Useful for navigating those relative links with ease." I consider this analogous to using `cd` on a Unix system; all references are relative unless the destination starts with `/`.

Actually, the vexing nature of the way things work now is more apparent if you have multiple path segments:
```
>>> boltons.urlutils.URL('https://host/a/b/c').navigate('d')
URL('https://host/a/b/d')
>>> boltons.urlutils.URL('https://host/a/b/c').navigate('../d')
URL('https://host/a/d')
>>> boltons.urlutils.URL('https://host/a/b/c/').navigate('../d')
URL('https://host/a/b/d')
>>> boltons.urlutils.URL('https://host/a/b/c').navigate('#d')
URL('https://host/a/b/c#d')
```

What I find most surprising is that, without a trailing slash on the base, `navigate` with a non-fragment argument _assumes_ the last part of the path is (erm, not sure about the right terminology) a "leaf" instead of a "stem" (or "file" instead of a "directory"?).

 To make a relative link from an URL without a trialing slash with the behavior you're describing, you have to: 1. `to_text()` 2. conditionally append a '/' (`normalize()` will not make `//` -> `/`) 3. recreate URL. Because the `URL` object is immutable, it's much easier to change the string argument to enable "upwards" navigation by pre-pending '../'.  This is what I would find less surprising:
```
>>> boltons.urlutils.URL('https://host/a/b/c').navigate('d')
URL('https://host/a/b/c/d')
>>> boltons.urlutils.URL('https://host/a/b/c').navigate('../d')
URL('https://host/a/b/d')
>>> boltons.urlutils.URL('https://host/a/b/c/').navigate('../d')
URL('https://host/a/b/d')
```

Unfortunately, I think this ship has sailed; changing this behavior would be massively backwards-incompatible.

PR #161, opened 2018-06-06
fixes files not been moved to destination when overwrite=False

uatach:


mahmoud:
Oh, good catch, thanks!

PR #162, opened 2018-06-19
Add MultiSub tool for single pass substitutions

induane:
MultiSub leverages regular expressions and the sub
module to perform a series of multiple replacements
in a string in a single pass.

For example:

long_string = long_string.replace('foo', 'bar')
long_string = long_string.replace('baz', 'fizz')
long_string = long_string.replace('bang', 'lame')

.... and so on - can get tedious and requires multiple
passes over the same string value.

The multisub equivilant would be:

m = MultiSub({
    'foo': 'bar',
    'baz': 'fizz',
    'bang': 'lame',
)
long_string = m.sub(long_string)

The longer the string in question the more of a
performance increase you'll see relative to a series
of calls to .replace

### DIFF
--- a/boltons/strutils.py
+++ b/boltons/strutils.py
@@ -1037,7 +1037,7 @@ class MultiSub(object):
             self.sub_data.append((exp, replacement))
 
         self.combined_pattern = re.compile('|'.join([
-            '(?:{})'.format(x.pattern) for x, _
+            '(?:{0})'.format(x.pattern) for x, _
             in self.sub_data
         ]))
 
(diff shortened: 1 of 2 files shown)


induane:
@mahmoud All pretty sweet suggestions. I've added a multi_replace function for simple invocation and added the regex options as well as an option to specify compile flags.

induane:
Made one more improvement; used named groups so we can figure out which replacement value to use automatically without needing to invoke the regex engine more.

mahmoud:
Great! Looks good to me, I'll merge it. One more idea I might implement before release is to just have a lower-cased convenience version, `multireplace(source_text, replacement_map)`, sort of like how the `AtomicSaver` has `atomic_save`. Thanks again for another welcome contribution :)

induane:
I think I made a lowercase convenience function already - but if I left it out then sorry about that. Thanks for merging this!

PR #174, opened 2018-10-10
Travis CI: Drop EOL Pythons, Upgrade to Py 3.7, 3.8-dev, and current PyPy

cclauss:
Upgrade to the _production_ version of Python 3.7 in alignment with travis-ci/travis-ci#9069

Also see the note about outdated Python versions at:
* https://docs.travis-ci.com/user/languages/python/#development-releases-support

Created #175 to highlight test failures on CPython 3.8.0.a0.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,22 +17,24 @@ python:
  
   # PyPy2.7: https://doc.pypy.org/en/latest
   # /index-of-release-notes.html#cpython-2-7-compatible-versions
-  - pypy         # Python 2.7.13 on PyPy 5.8.0
-  - pypy-6.0.0   # Python 2.7.13 on PyPy 6.0.0
+  - pypy          # Python 2.7.13 on PyPy 5.8.0
+  #- pypy-6.0.0   # Python 2.7.13 on PyPy 6.0.0  # Travis needs to upgrade!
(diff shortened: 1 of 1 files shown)


mahmoud:
Hey again @cclauss! Hmm, those 3.8 failures are concerning, I'll need to check on the changelog for the change in `dict` behavior and the `cgi` API removal. Thanks!

mahmoud:
Seems `cgi.escape` has been deprecated since 3.2 and is just being removed, that's easy enough to fix with a bit of copypasta.

As for the `dict` change, there are several changes, but the one from the [changelog](https://docs.python.org/dev/whatsnew/changelog.html#changelog) that seems closest is this: [Creating dict from OrderedDict doesn't preserve order](https://bugs.python.org/issue34320). I'll need to dig in a bit when it's less late. Thanks again!

cclauss:
Put Python 2.6 back into the testing.

Travis runs the __nightly__ build in __allow_failures mode_ so that we can get an idea of Python 3.8 issues but they do not fail the build.

mahmoud:
Oh sweet, thanks! OK, I'll create a separate ticket for tracking the `dict` changes.

PR #176, opened 2018-10-10
Tox and Appveryor: Upgrade to Python 3.7 and drop 2.6 

cclauss:
Python 2.6 was EOLed 5 years ago.  https://www.python.org/dev/peps/pep-0361/#release-lifespan

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -25,8 +25,7 @@ init:
 
 install:
   - "%PYTHON%/Scripts/easy_install -U pip"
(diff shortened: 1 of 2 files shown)


mahmoud:
Hey again @cclauss! While 3.7 support is definitely welcome, I know of at least one large codebase stuck on 2.6 that uses boltons and might appreciate bugfixes. If there were a really valuable feature that required dropping support, or if 3.7 support required it, that would be one thing, but this is something else. Anyways, if only there were a PEP that could port the world. :)

PR #178, opened 2018-10-11
Fix bug in implementation of `timeutils.daterange`

crazymerlyn:
Currently, the `finished` function used in case of `stop` being None has a
different signature than the one used in case of `stop` being a date.
This commit remedies that and adds a test for the same.


mahmoud:
Excellent catch, thank you!

PR #183, opened 2018-10-14
Add a __str__ function to boltons.urlutils.URL

### LINKED ISSUE #182: Represent the URL when str(url)
When working with `boltons.urlutils.URL`, it would be nice for `str(url)` to generate a string version of the URL, the same way it does when executing `url.to_text()`.

As of now it returns `URL('...')`, which is what you would expect when executing `repr()` but not `str()` (as per python docs, str should return a nice human representation whereas repr a formal python expression).

A part from making URL more intuitive and pythonic to use, this small change would allow it to nicely integrate with other libraries like [requests](https://github.com/requests/requests), which [automatically performs `str()`](https://github.com/requests/requests/pull/766) to non-string passed-in objects, so this would work:
```python
url = boltons.urlutils.URL('http://foo.com')
response = requests.get(url)
```
Which is just extra nice and useful :-)

Anyway, it is just a small enhancement in an already amazing library. Congratulations for it!

crazymerlyn:
This adds a `__str__` (and `__unicode__`) function to URL class which
returns a str representation of the url like `to_text()`.
Fixes #182.

### DIFF
--- a/boltons/urlutils.py
+++ b/boltons/urlutils.py
@@ -770,6 +770,12 @@ class URL(object):
         cn = self.__class__.__name__
         return u'%s(%r)' % (cn, self.to_text())
 
+    def __str__(self):
+        return self.to_text()
+
+    def __unicode__(self):
+        return self.to_text()
+
     def __eq__(self, other):
         for attr in self._cmp_attrs:
             if not getattr(self, attr) == getattr(other, attr, None):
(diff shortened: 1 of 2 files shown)


mahmoud:
Hehe, I mostly use https://github.com/python-hyper/hyperlink these days but sure :) Thank you!

PR #184, opened 2018-10-17
Support abstract @cachedproperty and @cachedmethods

mwilbz:
I find myself frequently using this sort of pattern:

```
class BaseMachineLearningModel(ABC):
    @cachedproperty
    @abstractmethod
    def load_big_model_resources(self):
        pass
```

While caching doesn't help the abstract method directly, it's helpful for communicating to IDEs, type checkers, and users of this class that implementations of this functions will be

1. properties
2. only executed once, even if called multiple times

However, if you call `BaseMachineLearningModel()`, we don't get a `TypeError` like you'd expect from an abstract class! This is because `@cachedproperty` doesn't maintain the `__isabstractmethod__` attribute of the function it wraps. This PR makes small modifications to make sure we maintain that attribute.

### DIFF
--- a/tests/test_cacheutils.py
+++ b/tests/test_cacheutils.py
@@ -1,6 +1,9 @@
 # -*- coding: utf-8 -*-
 
 import string
+from abc import abstractmethod, ABC
+
+import pytest
 
 from boltons.cacheutils import LRU, LRI, cached, cachedmethod, cachedproperty, MinIDMap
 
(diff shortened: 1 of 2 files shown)


tests/test_cacheutils.py
```diff
@@ -1,6 +1,9 @@
 # -*- coding: utf-8 -*-
-
+import abc
 import string
+from abc import abstractmethod, ABCMeta
```
wcooley: You only use `abc.ABCMeta`, so importing `ABCMeta` here is unnecessary.  Alternatively, use unqualified `ABCMeta` and remove the unnecessary `import abc`.

```suggestion
from abc import abstractmethod
```

mwilbz: Good catch! Removed the unnecessary import, thank you. =)

PR #189, opened 2018-11-22
Document that fileutils.iter_find_files works recursively on the spec…

Gallaecio:
…ified directory


mahmoud:
Good point, thanks!

PR #192, opened 2018-12-11
Travis CI has depreciated the use of 'sudo'

cclauss:
[Travis are now recommending removing __sudo__](https://blog.travis-ci.com/2018-11-19-required-linux-infrastructure-migration)


mahmoud:
Builds look green and in this case that's literally all that matters :) Thanks much!

Issue #194, opened 2019-01-15
wraps loses coroutinefunction-ness.


hynek:
I was thinking to replace wrapt with boltons in prometheus_async but I ran into a blocker:

```python
import asyncio

from boltons.funcutils import wraps


def delayed(func):
    @wraps(func)
    async def wrapped(*args, **kw):
        await asyncio.sleep(1.0)
        return await func(*args, **kw)

    return wrapped


async def f():
    await asyncio.sleep(0.1)


f2 = delayed(f)

print("f", asyncio.iscoroutinefunction(f))
print("f2", asyncio.iscoroutinefunction(f2))
```

prints:

```
f True
f2 False
```

At least on some versions of Python this is a problem, because the asyncio debug mode explodes.

wrapt and current versions of decorator.py handle it correctly.

hynek:
How often per day would you like to be pestered? ;)

PR #195, opened 2019-01-17
funcutils.wraps: Async coroutine support

mahmoud:
Here's my initial approach at fixing #194 for Py3.5+. Care to take a look @hynek?


hynek:
There is something not quite right.

```python
def yolo():
    def make_time_decorator(wrapped):
        @wraps(wrapped)
        async def decorator(*args, **kw):
            return (await wrapped(*args, **kw))
        return decorator

    return make_time_decorator


@yolo()
async def foo(x):
    await asyncio.sleep(x)


asyncio.run(foo(5))
```

works with the old version but in the new one, `decorator` is never awaited.

mahmoud:
Ah, I had a suspicion I was forgetting something. Fix pushed.

hynek:
Can’t test rn but I’d suggest to sleep less and use perf_counter or even perf_counter_ns (since it’s 3.7 anyway) to avoid a flaky slow test.

mahmoud:
Hmm, I haven't used it before, but I think perf_counter mostly gives us more resolution, but wouldn't really get rid of exposure to travis pauses.

Anyways, waiting 0.3 seconds won't really break the bank, it's still faster than the gc tests.

I'm fairly confident it's correct now, as I've just checked decorator.py and I landed on a very similar approach to what apparently works over there. :)

hynek:
More importantly it’s monotonic.

mahmoud:
haha, now that's a timely point I can get behind! fixed :)

mahmoud:
prepping for a release, you can try this out in 19.0 ;)

PR #196, opened 2019-01-28
Fix warnings in python3.7

crazymerlyn:
Importing Abstract Base Classes from `collections` instead of `collections.abc` is deprecated and will stop working in python 3.8. This modifies all such instances in the code to import from `collections.abc` first and only use `collections` as a fallback for older versions.


mahmoud:
Great! Thanks!

PR #199, opened 2019-01-30
dictutils: Add ManyToMany to go along with OneToOne

kurtbrose:


mahmoud:
Very nice! We'll need to document this, but docs are easy to incrementally add :)

PR #200, opened 2019-02-02
Fix issues with LRI Cache #155

CameronCairns:
Adds failing tests to exercise issues with LRI

To fix the issue, I have a trivial solution that has LRI inherit from LRU.  After a couple hours of thinking about it I don't think there's a faster solution that using the linked list that LRU uses without introducing an unbounded memory size of the underlying datastructure stored.

That being said LRU has a bunch of thread locking logic that I'm uncertain you would want to add to the LRI cache.  So the current solution is more of a conversation piece. Open to feedback on how to go forward on this:
1. Extract common logic from LRU and have LRI and LRU utilize that? (lets us omit thread safety from LRI if we want)
2. Keep naive solution (though thread locking might introduce unwanted overhead).
3. Maybe make thread safety an optional feature on both LRU and LRI. Would be tricky not to break existing features. Probably would require names like: LRU, ThreadUnsafeLRU, LRI, ThreadSafeLRI.

### DIFF
--- a/boltons/cacheutils.py
+++ b/boltons/cacheutils.py
@@ -312,6 +312,33 @@ class LRU(dict):
         return ('%s(max_size=%r, on_miss=%r, values=%s)'
                 % (cn, self.max_size, self.on_miss, val_map))
 
+class InheritingLRI(LRU):
+    """
+    The LRI is similar to a LRU except that we don't take
+    into consideration when an object was last inserted
+
+    The above class is used as a comparison of performance to
+    any LRI I try to build myself.  Unfortunately it requires
+    knowing quite a lot about the internals of LRU to work,
+    should the implementation details of LRU change, LRI
+    has a dependency on __getitem__ being the crux of the
+    access part of the algorith, and that the __setitem__
+    continues to treat inserts as "use" and move the key
+    to the front of the cache
+    """
+    def __getitem__(self, key):
+        with self._lock:
+            try:
+                link = self._link_lookup[key]
+            except KeyError:
+                self.miss_count += 1
+                if not self.on_miss:
+                    raise
+                ret = self[key] = self.on_miss(key)
+                return ret
+
+            self.hit_count += 1
+            return link[VALUE]
 
 class LRI(dict):
     """The ``LRI`` implements the basic *Least Recently Inserted* strategy to
--- a/boltons/cacheutils.py
+++ b/boltons/cacheutils.py
@@ -70,17 +70,16 @@ PREV, NEXT, KEY, VALUE = range(4)   # names for the link fields
 DEFAULT_MAX_SIZE = 128
 
 
-class LRU(dict):
-    """The ``LRU`` is :class:`dict` subtype implementation of the
-    *Least-Recently Used* caching strategy.
+class LRI(dict):
+    """The ``LRI`` implements the basic *Least Recently Inserted* strategy to
+    caching. One could also think of this as a ``SizeLimitedDefaultDict``.
 
-    Args:
-        max_size (int): Max number of items to cache. Defaults to ``128``.
-        values (iterable): Initial values for the cache. Defaults to ``None``.
-        on_miss (callable): a callable which accepts a single argument, the
-            key not present in the cache, and returns the value to be cached.
+    *on_miss* is a callable that accepts the missing key (as opposed
+    to :class:`collections.defaultdict`'s "default_factory", which
+    accepts no arguments.) Also note that, like the :class:`LRI`,
+    the ``LRI`` is instrumented with statistics tracking.
 
-    >>> cap_cache = LRU(max_size=2)
+    >>> cap_cache = LRI(max_size=2)
     >>> cap_cache['a'], cap_cache['b'] = 'A', 'B'
     >>> from pprint import pprint as pp
     >>> pp(dict(cap_cache))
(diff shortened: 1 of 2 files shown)


boltons/cacheutils.py
```diff
@@ -312,6 +312,33 @@ def __repr__(self):
...
+    The above class is used as a comparison of performance to
+    any LRI I try to build myself.  Unfortunately it requires
+    knowing quite a lot about the internals of LRU to work,
+    should the implementation details of LRU change, LRI
+    has a dependency on __getitem__ being the crux of the
+    access part of the algorith, and that the __setitem__
+    continues to treat inserts as "use" and move the key
+    to the front of the cache
+    """
+    def __getitem__(self, key):
```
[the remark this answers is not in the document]
mahmoud: That sounds right to me!

mahmoud:
I think the thread stuff is probably fine. I've kinda come around to the idea that it's better to be threadsafe than threadsorry on these sorts of datastructures. We can look at a new low-level cache with a different scheme for a future release. Did you take a look at how big the performance delta was with the inheriting LRI?

mahmoud:
Also, @kurtbrose might have some opinions on this, too.

CameronCairns:
Here are my attempts to profile the performance:
Setting Value with Eviction Speed
```
cache = LRI(1)
timeit.repeat("cache[1] = 1", "from __main__ import cache", 1000000)
```
LRI: `[0.6613960939994286, 0.6566329239994957, 0.6564849200003664, 0.6563216359991202, 0.6563244559984014]`
InheritingLRI: `[0.850525521000236, 0.845593550000558, 0.8448513969997293, 0.841510897000262, 0.8376139500014688]`

Setting Value without Eviction Speed
```
test_code = """
cache = cacheutils.LRI(1000)
for i in range(1000):
    cache[i] = i
"""
timeit.repeat(test_code, "from __main__ import cacheutils", 1000)
LRI: `[0.5287125860013475, 0.519335664999744, 0.519188601998394, 0.519156483000188, 0.5190456340005767]`
InheritingLRI: `[1.4676863400000002, 1.4665655310000147, 1.472885031998885, 1.4711705240006268, 1.4693267849997937]`
```

For reference the time to construct the datastructure is:
```
timeit.repeat("cache = cacheutils.LRI(1000)", 'from __main__ import cacheutils', number=1000) 
```
LRI: `[0.004028500999993412, 0.001178304000859498, 0.0011596250005823094, 0.001163509999969392, 0.0013013619991397718]`
InheritingLRI:` [0.004473193999729119, 0.0014076219995331485, 0.0013449090001813602, 0.001350635000562761, 0.0012909769993711961]`

So it's definitely slower, question is whether the slowness is worth the correctness.  I'd argue keeping around a dictionary of counts will only help you so much cause in degenerate cases the size of your deque will far outstrip the number of keys stored (so the time to find something to actually evict could take quite a while).

CameronCairns:
As suggested offline, here are my measurements around accessing values between the two:

```
cache = cacheutils.LRI(1000)
for i in range(1000):
 cache[i] = i
test_code = """
for i in range(1000):
    cache[i]
"""
timeit.repeat(test_code, "from __main__ import cache", 1000)
```
LRI: `[0.41686980899976334, 0.4089471110019076, 0.40890485199997784, 0.40919868299897644, 0.4087995760019112]`
InheritingLRI: `[0.5116791740001645, 0.5015979079980752, 0.5043692690014723, 0.5036118460011494, 0.5051938270007668]`

tests/test_cacheutils.py
```diff
@@ -25,11 +26,85 @@ def test_lru_add():
...
+
+
+def test_cache_sizes_on_repeat_insertions():
+    """
+    Regression test
+    Original LRI implementation had an unbounded size of memory
+    regardless of the value for it's `max_size` parameter due to a naive
+    insertion algorithm onto an underlying deque data structure. To
+    prevent memory leaks, this test will assert that a cache does not
+    grow past it's max size given values of a uniform memory footprint
```
mahmoud: its ;)

tests/test_cacheutils.py
```diff
@@ -25,11 +26,85 @@ def test_lru_add():
...
+        # sys.getsizeof object does not fully capture memory footprint
+        initial_key_sizes = {
+            k: sys.getsizeof(test_cache.__dict__[k])
+            for k in test_cache.__dict__
+        }
+        for k in test_cache:
+            for __ in range(100):
+                test_cache[k] = "1"
+        key_sizes_after_inserts = {
+            k: sys.getsizeof(test_cache.__dict__[k])
```
mahmoud: could we use `len()` instead of `sys.getsizeof()` here? PyPy doesn't implement getsizeof.

Also, if you can convert the dict comp to a list comp (`dict([(k, v) for k in ...])`), py26 should pass CI, too.

mahmoud:
Per offline discussion, I think if there's any change to the code itself it would be to make LRU inherit from LRI. But no obligation, if you'd like to merge it once the CI-fixing changes are made, that'd be perfect, too :)

tests/test_cacheutils.py
CameronCairns: woops

mahmoud:
All checks passing, merging! Thanks for all your hard work!

PR #202, opened 2019-02-10
FrozenDict

mahmoud:
Per your request @kurtbrose ;)

### DIFF
--- a/boltons/dictutils.py
+++ b/boltons/dictutils.py
@@ -966,4 +966,47 @@ def subdict(d, keep=None, drop=None):
 
     return dict([(k, v) for k, v in d.items() if k in keys])
 
+
+class FrozenDict(dict):
+    _hash = None
+
+    def updated(self, *a, **kw):
+        """Make a copy and add items from a dictionary or iterable (and/or
+        keyword arguments), overwriting values under an existing
+        key. See :meth:`dict.update` for more details.
+        """
+        data = dict(self)
+        data.update(*a, **kw)
+        return type(self)(data)
+
+    @classmethod
+    def fromkeys(cls, keys, value=None):
+        # one of the lesser known and used/useful dict methods
+        return cls(dict.fromkeys(keys, value))
+
+    def __repr__(self):
+        cn = self.__class__.__name__
+        return '%s(%s)' % (cn, dict.__repr__(self))
+
+    def __reduce_ex__(self, protocol):
+        return type(self), (dict(self),)
+
+    def __hash__(self):
+        if self._hash is None:
+            self._hash = hash(frozenset(self.items()))
+        return self._hash
+
+    def __copy__(self):
+        return self  # immutable types don't copy, see tuple's behavior
+
+    # block everything else
+    def _raise_frozen_typeerror(self, *a, **kw):
+        raise TypeError('%s object is immutable' % self.__class__.__name__)
+
+    __setitem__ = __delitem__ = update = _raise_frozen_typeerror
+    setdefault = pop = popitem = clear = _raise_frozen_typeerror
+
+    del _raise_frozen_typeerror
+
+
 # end dictutils.py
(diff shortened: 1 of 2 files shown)


boltons/dictutils.py
```diff
@@ -964,6 +964,49 @@ def subdict(d, keep=None, drop=None):
 
     keys = set(keep) - set(drop)
 
-    return dict([(k, v) for k, v in d.items() if k in keys])
+    return type(d)([(k, v) for k, v in d.items() if k in keys])
+
+
+class FrozenDict(dict):
+    _hash = None
```
kurtbrose: love the brevity -- could use `__slots__ = ('_hash',)` I think

boltons/dictutils.py
```diff
@@ -964,6 +964,49 @@ def subdict(d, keep=None, drop=None):
...
+    def __repr__(self):
+        cn = self.__class__.__name__
+        return '%s(%s)' % (cn, dict.__repr__(self))
+
+    def __reduce_ex__(self, protocol):
+        return type(self), (dict(self),)
+
+    def __hash__(self):
+        if self._hash is None:
+            self._hash = hash(frozenset(self.items()))
```
kurtbrose: this flexes up to 3x memory usage temporarily right?

I guess in python3 items() is a generator so only 2x

the other implementation used an accumulator that xor'd together hashes of keys + values

as long as you use a well behaved operator (like xor or sum-modulo-2**31) the order in which the values are accumulated together doesn't matter

on the other hand, frozenset.__hash__ is in C so is probably a lot faster; and frozenset pointed at the same keys + values is not doing copies of the underlying python objects, so the memory usage isn't really 2x/3x

boltons/dictutils.py
```diff
@@ -964,6 +964,49 @@ def subdict(d, keep=None, drop=None):
...
+        return self._hash
```
kurtbrose: maybe it should cache the negative result in _hash too? `try: except Exception as e: self._hash = e; raise` or similar?

boltons/dictutils.py
```diff
@@ -964,6 +964,49 @@ def subdict(d, keep=None, drop=None):
 
     keys = set(keep) - set(drop)
 
-    return dict([(k, v) for k, v in d.items() if k in keys])
+    return type(d)([(k, v) for k, v in d.items() if k in keys])
+
+
+class FrozenDict(dict):
```
kurtbrose: https://www.python.org/dev/peps/pep-0416/ linking to the rejected PEP from a docstring would probably be helpful for discoverability

it also links to a BUNCH of implementations to the point that it even categorizes them into whitelist/blacklist

the PEP points out one issue with the inherit-from-dict is that you can use `dict.__setitem__(frozen, key, val)` to "sneak around" the immutability

I'm not sure if that is really an issue, or maybe something worth documenting

kurtbrose:
looks pretty good, just some minor comments

kurtbrose:
:-p how long before we are doing combos -- OrderedFrozenOneToOne

boltons/dictutils.py
mahmoud: done!

boltons/dictutils.py
mahmoud: done!

mahmoud:
@kurtbrose could always turn it into a mixin! ;)

boltons/dictutils.py
mahmoud: Haha of course there'd be a PEP about this. Not sure looking at it would have sped things up though, probably would have just read a whole bunch of conflicting approaches instead of taking my own :P

Also, this is Python and there's lots of ways to sneak around. No point in chasing them all down. But also why document it when it's probably just going to enable it? :P

boltons/dictutils.py
mahmoud: Yeah, most of that occurred to me, but I figured I'd go for brevity and something I knew would be correct. The hash is cached, that's the main thing. Later, if you or I feel like optimizing we can always get in there and measure and such.

PR #203, opened 2019-02-12
Fix annotations that would otherwise blow up

### LINKED ISSUE #133: `boltons.funcutils.wraps` breaks with Python 3 function annotations

itamarst:
Fixes #133.

Thrown together pretty quickly, so review carefully in case I missed something.

### DIFF
--- a/boltons/funcutils.py
+++ b/boltons/funcutils.py
@@ -496,7 +496,7 @@ class FunctionBuilder(object):
     # def get_argspec(self):  # TODO
 
     if _IS_PY2:
-        def get_sig_str(self, stringify_annotations=False):
+        def get_sig_str(self, with_annotations=True):
             return inspect.formatargspec(self.args, self.varargs,
                                          self.varkw, [])
 
@@ -504,13 +504,11 @@ class FunctionBuilder(object):
             return inspect.formatargspec(self.args, self.varargs,
                                          self.varkw, [])[1:-1]
     else:
-        def get_sig_str(self, stringify_annotations=False):
-            if stringify_annotations:
-                annotations = {}
-                for key, value in self.annotations.items():
-                    annotations[key] = str(value)
-            else:
+        def get_sig_str(self, with_annotations=True):
+            if with_annotations:
                 annotations = self.annotations
+            else:
+                annotations = {}
             return inspect.formatargspec(self.args,
                                          self.varargs,
                                          self.varkw,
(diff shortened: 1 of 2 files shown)


boltons/funcutils.py
```diff
@@ -495,22 +496,26 @@ def __init__(self, name, **kw):
     # def get_argspec(self):  # TODO
 
     if _IS_PY2:
-        def get_sig_str(self):
+        def get_sig_str(self, with_annotations=True):
```
mahmoud: not sure if this flag is all that helpful here. not used in the body, and there's no argument for it in inspect.formatargspec on py2

mahmoud: if the goal is to make it forward compatible for py2/3 users, that's fine, maybe just a docstring letting them know it's ignored on py2 :)

itamarst: Went with removing it, since the flag is really only useful for internal implementation reasons, no reason for external caller to ever set it.

itamarst:
Thanks for the quick review, BTW. Any chance for a new release? :grin:

itamarst:
Ugh, removing flag in Python 2 version fails tests, cause it does get passed in. Docstring it is then.

mahmoud:
For you, @itamarst, anything ;)

https://pypi.org/project/boltons/19.0.1/

itamarst:
Thank you!

PR #205, opened 2019-02-13
Support for float priorities in priority queues

### LINKED ISSUE #204: Priority queue does not support float priorities
Through a lot of painful debugging in a larger codebase, I found that both of boltons priority queue implementations treat a float priority `p` as if it were `int(p)`.

```
In [1]: from boltons.queueutils import PriorityQueue

In [2]: items = [('a', 0.8), ('b', 0.9), ('c', 0.6), ('d', 1.8), ('e', 1.9), ('f', 1.6)]

In [3]: q = PriorityQueue()

In [4]: for i, p in items: q.add(i, p)

In [5]: for i in range(6): print(q.pop())
d
e
f
a
b
c
```

See the behavior that personally I would expect in this situation, exhibited by a depq package for double-ended priority queues:
```
In [6]: from depq import DEPQ

In [7]: q = DEPQ(items)

In [8]: for i in range(6): print(q.popfirst())
('e', 1.9)
('d', 1.8)
('f', 1.6)
('b', 0.9)
('a', 0.8)
('c', 0.6)
```

I believe the current behavior should be explicitly documented. It's not clear that the priorities would be treated as if they were floored neither from the current docs nor from a brief look at the source.

bogdan-kulynych:
Fix #204

### DIFF
--- a/boltons/queueutils.py
+++ b/boltons/queueutils.py
@@ -68,12 +68,12 @@ class BasePriorityQueue(object):
 
     Args:
         priority_key (callable): A function that takes *priority* as
-            passed in by :meth:`add` and returns an integer
+            passed in by :meth:`add` and returns a real number
             representing the effective priority.
 
     """
     # negating priority means larger numbers = higher priority
-    _default_priority_key = staticmethod(lambda p: -int(p or 0))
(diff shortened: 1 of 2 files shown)


mahmoud:
Excellent, thank you!

PR #210, opened 2019-05-07
added __rsub__ to complement-set

kurtbrose:
so that expressions of form set() - complement(set()) eval correctly


mahmoud:
Looks great, thanks!

PR #211, opened 2019-05-09
Fix incorrect substitution of flatten_iter() and flatten() in docstring

mosbasik:
This PR just fixes an incorrect reference to `iterutils.flatten_iter()` in the `iterutils.flatten()` docstring.


mahmoud:
Good catch, thank you!

PR #212, opened 2019-05-13
Docstring: FunctionBuilder 'default' argument should be a tuple

davidcaron:
In the docstring, it's mentionned that `defaults` must be a dict. But an error is raised when we pass a dict to `defaults`:

```
  File ".../site-packages/boltons/funcutils.py", line 616, in get_func
    func.__defaults__ = self.defaults
TypeError: __defaults__ must be set to a tuple object
```

The description in this PR is copied from https://docs.python.org/3/reference/datamodel.html (search for __defaults__)

It's the same for python 2 and 3.


mahmoud:
CI had a hiccup for some reason that gave me pause originally, but on a rerun, this looks like it's ready to go, thanks!

PR #217, opened 2019-07-11
Fix formatargspec deprecation warning

jpoehnelt:
>boltons/funcutils.py:528: DeprecationWarning: `formatargspec` is deprecated since Python 3.5. Use `signature` and the `Signature` object directly

- [x] add some tests to prevent regressions
- [x] ~switch to signature~
- [x] inline formatargspec

### DIFF
--- a/boltons/funcutils.py
+++ b/boltons/funcutils.py
@@ -786,55 +786,57 @@ except ImportError:
                 setattr(cls, opname, opfunc)
         return cls
 
+if not _IS_PY2:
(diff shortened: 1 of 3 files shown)


mahmoud:
Oh wow, a big miss on my part here. I was facing the same issue and copied over SQLAlchemy's approach _after_ this PR came in. And it was essentially the same as this one! I've resolved the conflicts and am definitely merging this for those tests. A big, belated sorry and thank you! Next time I'm running into issues, I'll be checking PRs _first_. Thanks again!

PR #218, opened 2019-08-03
Fixed doctests

Erotemic:
Hi, 

I fixed 4 of bolton's doctests that were broken. For whatever reason, doctest wasn't running them. I'm exactly not sure why pytest + doctest didn't catch these errors, but the stdlib `doctests` is weird, so I don't question it. 

I caught these errors using [xdoctest](https://github.com/Erotemic/xdoctest), (its a module I wrote to improve upon and hopefully replace the builtin doctest). Because xdoctest uses a abstract-syntax-trees instead of regular expressions to "parse" doctests, xdoctest is able to handle parsing real context free expressions, which regex simply cannot do.

In fact doctest was only running 132 of the 137 doctests run via xdoctest. 

I understand this lib wants 0 external deps, but if you used xdoctest as a testing requirement, you could easilly run these tests on the CI. I believe this project uses pytest, so I imagine a testing dependency is not out of the question. Running `pip install xdoctest` installs the core module, the standalone executable, and a pytest plugin. By default it is non-intrusive and does nothing, to enable the pytest plugin run with --xdoctest-modules instead of --doctest-modules. Then you will see 137 tests run instead of 132. 

Note that if you decide to go with xdoctest, please wait until the 0.9.2 release. I had to make a minor rework to the parser to make two fixes for backwards compatibility.  (Strange how I found errors in my lib by looking for errors in yours. FOSS can be nice like this)

For now this is just a minor patch that ensures users wont looks at these examples and get different results. 

--- 
EDIT; 
Not sure why pypy and 2.6 are failing. It seems to be network related.

### DIFF
--- a/boltons/fileutils.py
+++ b/boltons/fileutils.py
@@ -163,7 +163,7 @@ class FilePerms(object):
         Here's an example that holds true on most systems:
 
         >>> import tempfile
-        >>> 'r' in FilePerms.from_path(tempfile.tempdir).user
+        >>> 'r' in FilePerms.from_path(tempfile.gettempdir()).user
         True
         """
         stat_res = os.stat(path)
--- a/boltons/statsutils.py
+++ b/boltons/statsutils.py
@@ -583,12 +583,12 @@ class Stats(object):
         allowing for simple visualization, even in console environments.
 
         >>> data = list(range(20)) + list(range(5, 15)) + [10]
-        >>> print(Stats(data).format_histogram())
-         0.0:  5 ################################
-         4.4:  8 ###################################################
-         8.9: 11 ######################################################################
-        13.3:  5 ################################
-        17.8:  2 #############
+        >>> print(Stats(data).format_histogram(width=30))
+         0.0:  5 #########
+         4.4:  8 ###############
+         8.9: 11 ####################
+        13.3:  5 #########
+        17.8:  2 ####
 
         In this histogram, five values are between 0.0 and 4.4, eight
         are between 4.4 and 8.9, and two values lie between 17.8 and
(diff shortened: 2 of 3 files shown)


mahmoud:
Hi Jon,

First off, great job spotting those missed doctests. xdoctest looks solid and hopefully this issue helps demonstrate that value to those that'll need convincing that doctest could use a replacement.

As far as the test failures, it seems like Travis has been rolling out a new default image recently, with a new linux distro, and that new image/distro [doesn't have 2.6 support](https://travis-ci.community/t/issue-with-python-2-6-on-linux/3861). We could fix by pinning to trusty (14.04), but that's EOL, so I'm willing to say this spells the end for 2.6. The reasoning has always been 2.6 will be supported as long as it's easy to do so, and now it's looking like those days are over.

As far as PyPy goes, I'm guessing the new distro brought a new PyPy, too. That particular error is making itself look like a PyPy bug, but locally, on `pypy Python 2.7.13 (7.0.0+dfsg-1~ppa1~ubuntu16.04, Feb 07 2019, 20:56:09)`, everything's green. So I'll have to handle this out of band, too.

Anyways, TLDR not your problem, this is a great PR and I'll merge shortly :)

mahmoud:
Oh, if you'd like to add xdoctest feel free. I don't mind early-adopting new test deps as long as they're useful and the maintainer is committed to supporting them. :)

mahmoud:
PyPy issue tracked here: https://bitbucket.org/pypy/pypy/issues/3049/systemerror-in-rsocket__setblocking-travis

mahmoud:
Took me long enough! Thanks again!

PR #222, opened 2019-09-04
Adds attribute getter convenience to bucketize/partition

CameronCairns:
Bucketing/partitioning is often done based upon attribute access, so the
ability to pass in the attribute string should reduce the need for
boilerplate lambda functions. iterutils.unique was used as the template
for this change.

### DIFF
--- a/boltons/iterutils.py
+++ b/boltons/iterutils.py
@@ -471,15 +471,19 @@ def backoff_iter(start, stop, count=None, factor=2.0, jitter=False):
     return
 
 
-def bucketize(src, key=None, value_transform=None, key_filter=None):
-    """Group values in the *src* iterable by the value returned by *key*,
-    which defaults to :class:`bool`, grouping values by truthiness.
+def bucketize(src, key=bool, value_transform=None, key_filter=None):
+    """Group values in the *src* iterable by the value returned by *key*.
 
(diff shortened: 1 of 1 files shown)


boltons/iterutils.py
```diff
@@ -471,16 +471,20 @@ def backoff_iter(start, stop, count=None, factor=2.0, jitter=False):
     return
 
 
-def bucketize(src, key=None, value_transform=None, key_filter=None):
-    """Group values in the *src* iterable by the value returned by *key*,
-    which defaults to :class:`bool`, grouping values by truthiness.
+def bucketize(src, key=bool, value_transform=None, key_filter=None):
```
CameronCairns: What do you think about `None` -> `bool`?  It required that I modify partition but I do think it's a bit more self documenting. Let me know

mahmoud: Hmm, trying to remember why I did it that way, all those years ago. There has to be a better reason than "it looked better in the autogenerated signature docs", right? :P

Ultimately this seems like it does the same thing, so, I'm tentatively for it.

CameronCairns:
let me know if you need anything changed!

mahmoud:
Looks good to me, thanks again!

PR #223, opened 2019-09-19
Fix minor docstring formatting issues

connorbrinton:
Fixes include:
* `itertools.bucketize`:
  * Annotate `bool` reference
  * Display `key=` usage example as code block
* `setutils.complement`:
  * Display entire `NamesFilter` example as code block
  * Remove backticks from bad `NamesFilter` usage example


mahmoud:
Hey Connor! Thanks for the great contrib (as well as your patience with my reply and the dust of a PyPy bug on TravisCI). Looks good to me, merging!

PR #228, opened 2019-11-06
Port util_path from ubelt

Erotemic:
In #220 you mentioned `compressuser`, which I've renamed to `shrinkuser` (when I looked up "opposite of expand" in google, shrink was the first answer). I also selected several other related functions from ubelt.util_path and I omitted ones that had less usage.  

The 4 new functions introduced with brief justifications for their inclusion are:

* `shrinkuser` - the inverse of `os.path.expanduser`. Sometimes you want to transform a path to be relative to the user directory on any machine. Its also a more concise representation. 

* `expandpath` - os.path has `expanduser` and `expandvars`, but should a user really need to call two functions to expand strings that might start with `~` or `$HOME`? There should be a single function for tat. 

* `userhome` - The directory of the user directory. I'm not sure how useful this is by itself, but it is used to implement `shrinkuser` and it seems like a reasonable stdliby like thing.

* `augpath` - Using `join(spit(p)[0], splitext(basename(p))[0] + '.txt')` is really ugly, whereas `augpath(p, ext='.txt')` is an extremely concise way to specific modify parts of a filepath. This is a common task that should have a builtin function. I do think there is room for discussion on the name of the `multidot` keyword argument. I'm open to renaming that. I use this function often and every time I needed to set `mutltidot=True` it felt cumbersome. Maybe something like `dots=True` to indicate the extension may contain more than one dot?

Let me know what you think of the functionality and then I'll add CHANGELOG / other bookkeeping as necessary.

### DIFF
--- a/boltons/pathutils.py
+++ b/boltons/pathutils.py
@@ -71,11 +71,11 @@ def augpath(path, suffix='', prefix='', ext=None, base=None, dpath=None,
         >>> augpath('foo.bar', base='baz')
         'baz.bar'
         >>> augpath('foo.tar.gz', ext='.zip', multidot=True)
-        foo.zip
+        'foo.zip'
         >>> augpath('foo.tar.gz', ext='.zip', multidot=False)
-        foo.tar.zip
+        'foo.tar.zip'
         >>> augpath('foo.tar.gz', suffix='_new', multidot=True)
-        foo_new.tar.gz
+        'foo_new.tar.gz'
     """
     # Breakup path
     orig_dpath, fname = split(path)
--- a/boltons/pathutils.py
+++ b/boltons/pathutils.py
@@ -9,20 +9,15 @@ having to spend multiple lines of code splitting it up and stitching it back
 together.
 
 The :func:`shrinkuser` function replaces your home directory with a tilde.
-
-The :func:`userhome` function reports the home directory of the current user of
-the operating system.
 """
 from __future__ import print_function
 
-from os.path import (dirname, exists, expanduser, expandvars, join, normpath,
-                     split, splitext)
+from os.path import (expanduser, expandvars, join, normpath, split, splitext)
 import os
-import sys
 
 
 __all__ = [
-    'augpath', 'shrinkuser', 'userhome', 'expandpath',
+    'augpath', 'shrinkuser', 'expandpath',
 ]
 
 
@@ -127,7 +122,7 @@ def shrinkuser(path, home='~'):
         >>> assert shrinkuser(path + '/1', '$HOME') == join('$HOME', '1')
     """
     path = normpath(path)
-    userhome_dpath = userhome()
+    userhome_dpath = expanduser('~')
     if path.startswith(userhome_dpath):
         if len(path) == len(userhome_dpath):
             path = home
@@ -138,78 +133,20 @@ def shrinkuser(path, home='~'):
 
 def expandpath(path):
     """
-    Shell-like environment variable and tilde path expansion.
+    Shell-like expansion of environment variables and tilde home directory.
 
     Args:
-        path (str | PathLike): string representation of a path
+        path (str | PathLike): the path to expand
 
     Returns:
         str : expanded path
 
     Example:
-        >>> assert normpath(expandpath('~/foo')) == join(userhome(), 'foo')
+        >>> import os
+        >>> os.environ['SPAM'] = 'eggs'
+        >>> assert expandpath('~/$SPAM') == expanduser('~/eggs')
         >>> assert expandpath('foo') == 'foo'
     """
     path = expanduser(path)
     path = expandvars(path)
     return path
-
-
-def userhome(username=None):
-    """
-    Returns the path to some user's home directory.
-
-    Args:
-        username (str, default=None): name of a user on the system. If not
-            specified, the current user is inferred.
-
-    Returns:
-        str: userhome_dpath: path to the specified home directory
-
-    Raises:
-        KeyError: if the specified user does not exist on the system
-
-        OSError: if username is unspecified and the current user cannot be
-            inferred
-
-    Example:
-        >>> import getpass
-        >>> username = getpass.getuser()
-        >>> assert userhome() == expanduser('~')
-        >>> assert userhome(username) == expanduser('~')
-    """
-    if username is None:
-        # get home directory for the current user
-        if 'HOME' in os.environ:
-            userhome_dpath = os.environ['HOME']
-        else:  # nocover
-            if sys.platform.startswith('win32'):
-                # win32 fallback when HOME is not defined
-                if 'USERPROFILE' in os.environ:
-                    userhome_dpath = os.environ['USERPROFILE']
-                elif 'HOMEPATH' in os.environ:
-                    drive = os.environ.get('HOMEDRIVE', '')
-                    userhome_dpath = join(drive, os.environ['HOMEPATH'])
-                else:
-                    raise OSError("Cannot determine the user's home directory")
-            else:
-                # posix fallback when HOME is not defined
-                import pwd
-                userhome_dpath = pwd.getpwuid(os.getuid()).pw_dir
-        return userhome_dpath
-
-    # A specific user directory was requested
-    if sys.platform.startswith('win32'):  # nocover
-        # get the directory name for the current user
-        c_users = dirname(userhome())
-        userhome_dpath = join(c_users, username)
-        if not exists(userhome_dpath):
-            raise KeyError('Unknown user: {}'.format(username))
-    else:
-        import pwd
-        try:
-            pwent = pwd.getpwnam(username)
-        except KeyError:  # nocover
-            raise KeyError('Unknown user: {}'.format(username))
-        userhome_dpath = pwent.pw_dir
-    return userhome_dpath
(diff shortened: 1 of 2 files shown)


boltons/pathutils.py
```diff
@@ -0,0 +1,212 @@
...
+        elif path[len(userhome_dpath)] == os.path.sep:
+            path = home + path[len(userhome_dpath):]
+    return path
+
+
+def expandpath(path):
+    """
+    Wrapper around expanduser and expandvars.
+
+    Less aggressive than truepath. Only expands environs and tilde. Does not
```
mahmoud: Hmm, what's truepath?

boltons/pathutils.py
```diff
@@ -0,0 +1,212 @@
...
+from __future__ import print_function
+
+from os.path import dirname
+from os.path import exists
+from os.path import expanduser
+from os.path import expandvars
+from os.path import join
+from os.path import normpath
+from os.path import split
+from os.path import splitext
```
mahmoud: How about importing these in this style:

```python 
from os.path import (dirname, 
                     exists, 
                     expanduser, ...)`

File: boltons/pathutils.py [hunk 3]
```diff
@@ -0,0 +1,212 @@
...
+        >>> augpath('foo.bar', prefix='_')
+        '_foo.bar'
+        >>> augpath('foo.bar', base='baz')
+        'baz.bar'
+        >>> augpath('foo.tar.gz', ext='.zip', multidot=True)
+        foo.zip
+        >>> augpath('foo.tar.gz', ext='.zip', multidot=False)
+        foo.tar.zip
+        >>> augpath('foo.tar.gz', suffix='_new', multidot=True)
+        foo_new.tar.gz
```
mahmoud: Ah, thanks for this example. I was wondering what `multidot` meant practically, and I see it now.

boltons/pathutils.py
```diff
@@ -0,0 +1,212 @@
...
+        base = orig_base
+    # Recombine into new path
+    new_fname = ''.join((prefix, base, suffix, ext))
+    newpath = join(dpath, new_fname)
+    return newpath
+
+
+def shrinkuser(path, home='~'):
+    """
+    Inverse of ````os.path.expanduser````
```
mahmoud: How about linking to the builtin docs, like:

```
:func:`os.path.expanduser`
```

boltons/pathutils.py
```diff
@@ -0,0 +1,212 @@
...
+    and/or suffix.
+
+    A prefix is inserted before the basename. A suffix is inserted
+    between the basename and the extension. The basename and extension can be
+    replaced with a new one. Essentially a path is broken down into components
+    (dpath, base, ext), and then recombined as (dpath, prefix, base, suffix,
+    ext) after replacing any specified component.
+
+    Args:
+        path (PathLike): a path to augment
```
mahmoud: when we say PathLike, it means a path-like str, right?

boltons/pathutils.py
```diff
@@ -0,0 +1,212 @@
...
+        >>> assert expandpath('foo') == 'foo'
+    """
+    path = expanduser(path)
+    path = expandvars(path)
+    return path
+
+
+def userhome(username=None):
+    """
+    Returns the user's home directory.
```
mahmoud: Maybe this docstring should be expanded a bit to specify that a user's home directory can be different than `os.path.expanduser(`~user`)`, and this handles that complexity.

boltons/pathutils.py
```diff
@@ -0,0 +1,212 @@
...
+    Example:
+        >>> import getpass
+        >>> username = getpass.getuser()
+        >>> assert userhome() == expanduser('~')
+        >>> assert userhome(username) == expanduser('~')
+    """
+    if username is None:
+        # get home directory for the current user
+        if 'HOME' in os.environ:
+            userhome_dpath = os.environ['HOME']
```
mahmoud: Small thing, but I think I prefer this to be structured as a guard statement and just early return, so the big else block can be dedented.

Also, why not do `expanduser('~')` here?

boltons/pathutils.py
```diff
@@ -0,0 +1,212 @@
...
+
+The ``expandpath`` function expands the tilde to $HOME and environment variables
+to their values.
+
+The ``augpath`` function creates variants of an existing path without having to
+spend multiple lines of code spliting it up and stitching it back together.
+
+The ``shrinkuser`` function replaces your home directory with a tilde.
+
+The ``userhome`` function reports the home directory of the current user of the
```
mahmoud: If you do:
```
:func:`userhome`
```

The docs will have links to the functions below.

boltons/pathutils.py
Erotemic: I'm not sure what the best "type" to use here is. What I want to communicate is that path could be either a str or pathlib.Path object.

I ultimately started using PathLike in reference this: https://docs.python.org/3/library/os.html#os.PathLike but I'm not sure if its correct. Perhaps the proper type is actually `str | PathLike`?

boltons/pathutils.py
Erotemic: truepath is a function I opted not to port, and as such it should not be referenced here. I'll fix that.

boltons/pathutils.py
Erotemic: This is an artifact of my :AutoImport vim command, which automatically tries to add import statements for undefined names. I can refactor it to the concise style.

boltons/pathutils.py
Erotemic: I tend to prefer code that doesn't have early return statements. It makes it much easier to copy-paste segments into IPython when I'm doing interactive development. I also like having a single return as the last line of a function, so you know that whatever code path is taken must always populate that specific return variable. However, this is only a minor preference, and I'm willing to make this modification. 

However, I would prefer to keep the existing implementation without using `expanduser('~')`. The expanduser function is actually defined in different files in the stdlib depending on your OS. I think its nicer to have a single code path that shows behavior on all system. Also this code cuts out a lot of the extra work that `expanduser('~')` would have to do because we aren't replacing anything in a string we are only finding the path to the user's home directory. While using expanduser is more concise I think the current implementation demystifies what expanduser is actually doing under the hood. My vote is to keep this, but I'll yield to your opinion if you feel strongly about it.

boltons/pathutils.py
Erotemic: This is fixed.

boltons/pathutils.py
Erotemic: Do you have any ideas for a better name than `multidot`? I was thinking `dots` as a shorthand, but I'm not sure if it becomes unclear as to what its doing.

Or perhaps `ndots`, where you specify the maximum number of dots allowed in an extension?

boltons/pathutils.py
Erotemic: I ended up swapping the input args for `str | PathLike`, but I noticed that even when a `pathlib.Path` object is the argument the output type is always `str`, so I swapped the return types for that.

boltons/pathutils.py
Erotemic: I'm actually not even sure that is the case. I think the user's home directory will always be the same as `os.path.expanduser('~user')`. The only difference is that if `user` doesn't exist, userhome will raise an error, whereas expanduser will return `~user`. (Unless you saw something to indicate that's not the case, its been awhile since I wrote that function). 

Now that I'm thinking about it, I'm not sure how useful a `userhome` function is. It seems stdliby, but I've never actually used for anything except for writing `compressuser` and testing other functions in pathutils. One option is to simply remove it and user `expanduser('~')` in the one other place where it is used in this module. On the other hand, just because I don't use it doesn't mean someone else wont find it useful.

mahmoud: Hmm, I lean toward `expanduser('~')`, because that's what I've done in the past and I don't have a great desire to keep up with how OSes I'm not running are changing (unless there's really no maintained builtin).

Erotemic: Sounds good. I've removed `userhome` entirely.  Best to keep things simple.

boltons/pathutils.py
mahmoud: hmm, been thinking about this one. I can't remember whether I knew the number of dots back when I was splitting extensions last. I could see `ndots` or `maxdots` though. 

If it stays a bool flag, then the only names that pop into my head are `greedy` and `eager`. Multidot isn't that bad though.

Erotemic: Thanks for giving this some thought. I guess it make sense to leave it as a boolean with the current name, given that nothing comes to mind easily. 

The main issue I have with the name is that it is 8 characters plus the 5 characters for `=True`. I would prefer to have something slightly more compact because I often find that setting this flag myself going over the 80 character line limit when I use this function, but that is a first world problem. 

I think the other part that bugs me about this is that in general it's not possible to know whether text after the first dot is meant to be a file extension or is actually part of the file name. Sure there are heuristics (e.g. look for `.tar.gz`), but these break down easily (e.g. `.json.custom.backup`). I also agree that I typically don't know the exact number of dots that are possible I'm expecting when processing a filepath, but I typically will know if dots are allowed in the filename or not, so it might be better to opt for the simpler boolean API.

mahmoud:
Hey @Erotemic just saw that the last commit was "`wip`", wondering if there's more you had in mind, or if this is pretty much ready to go? I can help with the changelog when release day comes :)

Erotemic:
This is pretty much ready to go. The "wip" is because I'm often lazy with my commit messages (`alias gcwip='git commit -am "wip" && git push'`).  Let me know if there is anything I should do with changelog or other book-keeping.

mahmoud:
Excellent, merged! I'll try to do one last release in 2019, but otherwise this may be boltons==20.0.0! :)

PR #234, opened 2020-01-20
Implement "seekable" method as required by spec

### LINKED ISSUE #235: seekable not implemented on SpooledStringIO & SpooledBytesIO
Some libraries expect that a file like object should support a "seekable" method which returns a boolean and these crash when used with SpooledStringIO and SpooledBytesIO

See:
https://docs.python.org/2/library/io.html#io.IOBase.seekable
https://docs.python.org/3/library/io.html#io.IOBase.seekable

induane:
https://docs.python.org/2/library/io.html#io.IOBase.seekable
https://docs.python.org/3/library/io.html#io.IOBase.seekable

Seekable should be implemented as a method on file like
objects.

Fixes #235

### DIFF
--- a/boltons/ioutils.py
+++ b/boltons/ioutils.py
@@ -170,6 +170,12 @@ class SpooledIOBase(object):
     def seekable(self):
         return True
 
+    def readable(self):
+        return True
+
+    def writable(self):
+        return True
+
     __next__ = next
 
     def __len__(self):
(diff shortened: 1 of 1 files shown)


mahmoud:
This is good in and of itself, but by the same token, shouldn't we also have `readable()` and `writable()`?

induane:
Ahh good call.

mahmoud:
Now we're talkin, looks good! Thanks!

PR #244, opened 2020-03-18
Non-decorator version of funcutils.wraps and option to hide wrapped function.

### LINKED ISSUE #242: funcutils.wraps() not working with partials
Hi!

With boltons 20.0.0 (on Linux), I get errors when trying to wrap partials. My goal is to fix the docstring and signature of the wrapped partial, so it acts as a normal function, with an option to "hide" the given arguments with "injected".

Examples:
```
from functools import partial

def func(a, b=1, c=1):
    """Docstring"""
    return (a, b, c)

newf = wraps(func)(partial(func, b=2))
newf(1)
# Expected (1, 2, 1)
# Got : TypeError : func() got multiple values for argument 'b'
```

After inspection, I believe this is due to `partial` modifying the signature of `func` by changing `b` and  all subsequent  args to "keyword-only". Which `wraps` doesn't see as it builds the new function from `func` directly.

```
from inspect import signature

signature(partial(func, b=2)) # <Signature (a, *, b=2, c=1)>
signature(newf, follow_wrapped) # <Signature (a, b=1, c=1)>
``` 

So. First of all, is this a bug of `boltons.funcutils.wraps` or is this just not in the scope of what it is supposed to do?
I could use `functools.update_wrapper` but I really wanted to "inject" the new keywords, make them disappear from the signature. Also, using `help()` on a "updated" partial doesn't work, while it does on  `wraps(func)(partial(func))`.

I quickly modified `wraps` and was able to obtain the wanted behaviour by passing the partial func directly and using that one to create the underlying `FunctionBuilder` instance. It works since the arguments are the same, just with modified properties and defaults, but it kinda doesn't make sense since `wraps` is supposed to be a decorator. Thus a solution I see is to move the code from `wraps` to a new  `update_wrapper`. Example:

```
def update_wrapper(wrapper, wrapped, build_from=None, injected=None, expected=None, **kw):
     ... same code as `wraps` but:
     fb = FunctionBuilder.from_func(build_from or wrapped)
     ... and it returns "fully_wrapped"

def wraps(func, injected=None, expected=None, **kw):
    def wrapper_wrapper(wrapper_func):
        return update_wrapper(wrapper_func, func, injected=injected, expected=expected, **kw)
    return wrapper_wrapper
```

What do you think?
(If needed or wanted, I have time to suggest a PR on this)

aulemahal:
Hi!

As as discussed in #242, here is my suggestion to add an edge case to the usage of `funcutils.wraps`. 
The situation is that if the wrapped and wrapper function have the same arguments but differ as to which are keyword-only or positional-only, the function returned by `wraps` will fail as the wrong "invocation string" was inserted in the process. A solution to this is to create the internal `FunctionBuilder`instance from the wrapper instead of the wrapped. As it didn't make sense to simply add an argument to `wraps`, I decided to replicate the structure of the built-in `functools` and add a `update_wrapper` function, of which `wraps` is only a `partial` call.

This solves a problem that originated from using partials. Code that now works to wrap those is as follow:
```python
import functools
from boltons import funcutils

def func(a, b=1, c=1):  # Has the signature "a, b=1, c=1"
    return a, b, c

wrapper = partial(func, b=2)
functools.update_wrapper(wapper, func)  # Needed as FunctionBuilder will look for some attributes like __name__ that are not updated directly with `partial`.
# wrapper has the signature; "a, *, b=2, c=1"

new_func = funcutils.update_wrapper(wrapper, func, build_from=wrapper, injected='b')
# new_func now has the signature : "a, *, c=1"
new_func(1)  # Prints : 1, 2, 1, 
```

Also, another small improvement I made here is to add an option to completely hide the wrapped function in the new one returned. Most IDEs use `inspect.signature` to extract the signature functions that they display to the user. However, by default, `signature` follows all wrapped functions until the innermost, so the user doesn't see the updated signature. One could use the `follow_wrapped=False` option, but I believe offering the possibility here could be nice.

Closes #242

### DIFF
--- a/boltons/funcutils.py
+++ b/boltons/funcutils.py
@@ -536,9 +536,10 @@ def update_wrapper(wrapper, func, injected=None, expected=None, build_from=None,
             the wrapper (the opposite of *injected*). See
             :meth:`FunctionBuilder.add_arg()` for more details.
         build_from (function): The callable from which the new wrapper
-            is built. Defaults to *func*. Useful in some specific cases
-            where wrapper and func have the same arguments but differ on
-            which are keyword-only and positional-only.
+            is built. Defaults to *func*, unless *wrapper* is partial object
+            built from *func*, in which case it defaults to *wrapper*.
+            Useful in some specific cases where *wrapper* and *func* have the
+            same arguments but differ on which are keyword-only and positional-only.
         update_dict (bool): Whether to copy other, non-standard
             attributes of *func* over to the wrapper. Defaults to True.
         inject_to_varkw (bool): Ignore missing arguments when a
@@ -571,6 +572,9 @@ def update_wrapper(wrapper, func, injected=None, expected=None, build_from=None,
     hide_wrapped = kw.pop('hide_wrapped', False)
     if kw:
         raise TypeError('unexpected kwargs: %r' % kw.keys())
+
+    if isinstance(wrapper, functools.partial) and func is wrapper.func:
+        build_from = build_from or wrapper
 
     fb = FunctionBuilder.from_func(build_from or func)
 
--- a/boltons/funcutils.py
+++ b/boltons/funcutils.py
@@ -38,6 +38,11 @@ try:
 except ImportError:
     NO_DEFAULT = object()
 
+try:
+    from functools import partialmethod
+except ImportError:
+    partialmethod = None
+
 
 _IS_PY35 = sys.version_info >= (3, 5)
 if not _IS_PY35:
@@ -242,11 +247,14 @@ class InstancePartial(functools.partial):
     has the same ability, but is slightly more efficient.
 
     """
+    if partialmethod is not None:  # NB: See https://github.com/mahmoud/boltons/pull/244
+        @property
+        def _partialmethod(self):
+            return partialmethod(self.func, *self.args, **self.keywords)
+
     def __get__(self, obj, obj_type):
         return make_method(self, obj, obj_type)
 
-    def __set__(self, instance, value):
-        pass
 
 
 class CachedInstancePartial(functools.partial):
@@ -259,6 +267,11 @@ class CachedInstancePartial(functools.partial):
 
     See the :class:`InstancePartial` docstring for more details.
     """
+    if partialmethod is not None:  # NB: See https://github.com/mahmoud/boltons/pull/244
+        @property
+        def _partialmethod(self):
+            return partialmethod(self.func, *self.args, **self.keywords)
+
     def __get__(self, obj, obj_type):
         # These assignments could've been in __init__, but there was
         # no simple way to do it without breaking one of PyPy or Py3.
@@ -280,9 +293,6 @@ class CachedInstancePartial(functools.partial):
         except KeyError:
             obj.__dict__[name] = ret = make_method(self, obj, obj_type)
             return ret
-
-    def __set__(self, instance, value):
-        pass
 
 
 partial = CachedInstancePartial
--- a/boltons/funcutils.py
+++ b/boltons/funcutils.py
@@ -526,7 +526,11 @@ def update_wrapper(wrapper, func, injected=None, expected=None, build_from=None,
             attributes of *func* over to the wrapper. Defaults to True.
         inject_to_varkw (bool): Ignore missing arguments when a
             ``**kwargs``-type catch-all is present. Defaults to True.
+        hide_wrapped (bool): Remove reference to the wrapped function(s)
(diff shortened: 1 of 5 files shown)


mahmoud:
Nice! Is there some way to structure this so that we don't need to call `functools.update_wrapper()`?

aulemahal:
The best way would be to modify `FunctionBuilder.from_func` to circumvent the issue when encountering a `partial`! Or add a `FunctionBuilder.from_partial`. I didn't do it here, since you seemed not to be a fan, but it could be arranged.

Note: boltons' `funcutils.partial` still will not work. That's because the call to `inspect.getfullargspec()` will raise an error. It has to do with the added `__get__` method, but I'm not sure.

aulemahal:
The last two commit add support for partials in FunctionBuilder (and move py3 tests to py3 test files).
I decided to drop partials support in python2, I wasn't able to find a elegant way to do it.

Also, boltons' partials (InstancePartial and CachedInstancePartial) still did not work because inspect sees them as "method descriptors". My hack was to add a __set__ method that does nothing, so inspect sees them as "data descriptor" instead and goes on. It doesn't seem to interfere with the rest, as the tests all pass, but you may not want this kind of hacky solution.

mahmoud:
Hmm, the issue with using a data descriptor is that now reassignment of that attribute on the owning instance isn't possible, as it would be with methods (methods are data descriptors). I've confirmed this with a test (which I'll commit shortly), and right now I'm looking for another workaround. Thanks for the progress thus far!

mahmoud:
OK, spent a decent number of minutes rooting around inspect.py and here's what I've got so far:

- `signature()` has some special handling around partials, but those aren't hit with this, because
- `InstancePartial`, because it has a `__get__` without a `__set__` makes signature think it's a `methoddescriptor` (which it sort of is), but specifically `signature` thinks this makes it a "builtin". (This is the condition [right above](https://github.com/python/cpython/blob/9a81ab107a54b8ca320fb703f7c68e14ccd9d016/Lib/inspect.py#L2294) where `signature` would check if we were a partial)
- At this point I tried making the `InstancePartial`/`CachedInstancePartial` inherit from `functools.partialmethod` instead of `functools.partial`, but the behavior between the two is quite different.
- Then I noticed that there's `a special handshake for partialmethods`. Adding that seems to have fixed the issue, without resorting to losing our `methoddescriptor` status.

All this makes me a bit nervous, though. I might want to file a bug with CPython so this is on the maintainers' radar. 

PS  `partialmethod` is a real disappointment. I don't think this is the first time it's let me down. :/

aulemahal:
Interesting! Thanks for digging on that.

mahmoud:
I think I'm ok without supporting partials on Python 3. The more I think about it, the more strange and niche it seems. Efforts are probably best spent elsewhere.

One thing still troubles me. Looking at the test for wrapping:

```
@pytest.mark.parametrize('partial_kind', (functools, funcutils))
def test_update_wrapper_partial(partial_kind):
    wrapper = partial_kind.partial(wrappable_varkw_func, b=1)
    functools.update_wrapper(wrapper, wrapper.func)

    fully_wrapped = update_wrapper(wrapper, wrappable_varkw_func, build_from=wrapper)
    assert fully_wrapped(1) == (1, 1)
```

It still looks wrong to me that we'd need to a) call `functools.update_wrapper` then call `funcutils.update_wrapper`, and b) pass `wrapper` twice to `funcutils.update_wrapper` (once as the first arg, once as `build_from`). 

Can we make it so `funcutils.update_wrapper` subsumes the whole dance, and have the test setup consist only of `fully_wrapped = update_wrapper(wrapper, wrappable_varkw_func)`?

aulemahal:
Sorry, the first commit is just that I forgot to remove the first call to `functools.update_wrapper` in my test, my fix for partials did fix this thing.
However, for your comment b), I'm less sure about this, but the second commit adds a test in boltons' `update_wrapper` for the specific case of `wrapper` being a partial of `func` (our case in the test). In that case, it makes `build_from` default to wrapper instead of `func`.

mahmoud:
Seems like this one's pretty close, and the last two commits look good. Was there anything remaining from your end before merge?

aulemahal:
Everything is fine for me! Thanks!
When do you expect to publish the next version of boltons?

mahmoud:
Great. I was looking at doing that this weekend, if all loose ends are wrapped up. Will keep you posted :)

mahmoud:
This is now released in boltons 20.1.1. Thanks again, and enjoy!

PR #245, opened 2020-03-22
Fix singularize("senses") == "sens"

### LINKED ISSUE #241: singularize("senses") == "sens"
`strutils.singularize("senses") == "sens"`, but the correct singular is "sense". That's caused by this condition in `singularize`:

```python
elif word.endswith('es') and word[-3] == 's':
    singular = word[:-2]
```

I think "sense" might be irregular, because there are other very similar words for which `singularize()` works correctly, e.g. `singularize("lenses") == "lens"`.

orn688:
"senses" turns out to be irregular with respect to the current singularization algorithm.

Fixes #241


mahmoud:
Perfect! Thank you!

mahmoud:
This is now released in boltons 20.1.1. Thanks again, and enjoy!

PR #246, opened 2020-03-25
Fix NetstringSocket to correctly use arguments to .read_ns() method.

### LINKED ISSUE #243: Unused argument in NetstringSocket
Following line is incorrect and can be confusing:
https://github.com/mahmoud/boltons/blob/9c99fd6e4a345803a43003284cd2aed8a772eb01/boltons/socketutils.py#L640

The value of `maxsize` is never used by this method.

ikamensh:
fixes #243 - reproduced a problem with unit test and fixed it.


mahmoud:
Hey @ikamensh! This looks excellent! I believe the CI failure is just due to the netstring server thread blocking shutdown (appveyor took an hour before it killed the test process). I'd try making it a daemon thread for simplicity's sake. Otherwise tests pass, so as soon as that's resolved, I'm happy to merge.

Issue #122, opened 2017-03-22
strip() that works with iterables


kurtbrose:
I'd like strip / rstrip / lstrip that works with arbitrary iterables:

```python
def iter_lstrip(li, it): 
    pos = 0
    while li[pos] == it:
        pos += 1
        if pos == len(li):
            return []
    return li[pos:]


def iter_rstrip(li, it):
    return list(reversed(iter_lstrip(list(reversed(li)), it)))


def iter_strip(li, it):
    return iter_lstrip(iter_rstrip(li, it), it)
```

```python
iter_strip([0,0,0,1,2,3,4,0,0], 0)
# [1,2,3,4]
```

mahmoud:
Sounds good to me! However, in order to be `iter_` functions, i think they might have to be generators, and maybe there's something to be said for making these work with arbitrary iterables?

mblahay:
I have some code I can modify and then contribute to this. What module should this go into?

@mahmoud, So that I am clear, the criticism with the example provided by @kurtbrose is that it seems to focus on sequences (most probably a list in his example), rather than an arbitrary iterable, correct?

kurtbrose:
yes, basically I'd like to have the same behavior as the string functions strip, rstrip, lstrip that work over arbitrary iterables

maybe there are other string methods which would also be handy (translate?) -- just making the code more cohesive and readable b/c everybody is already familiar with the string methods

mblahay:
Here is an example of iter_lstrip:

`
def iter_lstrip(iterable, value):
    iterator = iter(iterable)
    try:
        #step through the start of iterator, looking for all matching 
        #items that match the value to be stripped
        i = next(iterator)
        while i == value:
            i = next(iterator)

        #yield all remaining items that exist after stripped values
        yield i
        yield from iterator

    #graceful handling of StopIteration within a generator
    except StopIteration:
        pass
`

mblahay:
def iter_lstrip(iterable, strip_value):
        iterator = iter(iterable)
        for i in iterator:
            if i != strip_value: #Check for end of strip_value series
                yield i
                break
        yield from iterator

mblahay:
Here is a function for the right side, but it is admittedly ugly. Suggestions to make it look better are welcome.

    def iter_rstrip(iterable,strip_value):
        iterator = iter(iterable)
        for i in iterator:
            if i == strip_value:
                l = list() #List is used here as a cache
                l.append(i)
                broken = False
                for i in iterator:
[...]

PR #253, opened 2020-04-21
#122 - Adding lstrip, rstrip, and strip for iterables

mblahay:
Adding lstrip, rstrip, and strip for iterables

### DIFF
--- a/boltons/iterutils.py
+++ b/boltons/iterutils.py
@@ -18,6 +18,7 @@ import random
 import socket
 import hashlib
 import itertools
+import sys
 
 try:
     from collections.abc import Mapping, Sequence, Set, ItemsView, Iterable
@@ -200,10 +201,11 @@ def lstrip_iter(iterable, strip_value=None):
     """    
     iterator = iter(iterable)
     for i in iterator:
-        if i != strip_value: #Check for end of strip_value series
+        if i != strip_value:
             yield i
             break
-    yield from iterator
+    for i in iterator:
+        yield i
 
 
 def rstrip(iterable, strip_value=None):
@@ -230,18 +232,19 @@ def rstrip_iter(iterable, strip_value=None):
     iterator = iter(iterable)
     for i in iterator:
         if i == strip_value:
-            l = list() #List is used here as a cache
-            l.append(i)
+            cache = list()
+            cache.append(i)
             broken = False
             for i in iterator:
                 if i == strip_value:
-                    l.append(i)
+                    cache.append(i)
                 else:
                     broken = True
                     break
-            if not broken:
-                return #Return to caller here because the end of the iterator has been reached
-            yield from l
+            if not broken: # Return to caller here because the end of the
+                return     # iterator has been reached
+            for t in cache:
+                yield t
         yield i
 
 
--- a/boltons/iterutils.py
+++ b/boltons/iterutils.py
@@ -253,7 +253,7 @@ def strip(iterable, strip_value=None):
     will match the value of the argument strip_value. Functionality is 
     analigous to that of the method str.strip. Returns a list.
 
-    >>> rstrip(['Fu', 'Foo', 'Bar', 'Bam', 'Fu'], 'Fu')
+    >>> strip(['Fu', 'Foo', 'Bar', 'Bam', 'Fu'], 'Fu')
     ['Foo', 'Bar', 'Bam']
 
     """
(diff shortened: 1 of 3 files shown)


mahmoud:
Cool, I'm generally +1 on this, but I'll defer to @kurtbrose somewhat since it was his idea. If he doesn't chime in, I'll come back through and give it a closer look. Thanks!

mblahay:
@mahmoud, I haven't heard from @kurtbrose nor do I see any recent activity on his account. How long do you want to wait?

mblahay:
@mahmoud, what is the status of this?

mblahay:
poke

mahmoud:
Yeah I don't know where he's at, busy presumably. I'll do a preliminary review.

boltons/iterutils.py
```diff
@@ -177,6 +178,100 @@ def split_iter(src, sep=None, maxsplit=None):
...
+def lstrip(iterable, strip_value=None):
+    """Strips values from the beginning of an iterable. Stripped items will
+    match the value of the argument strip_value. Functionality is analigous
+    to that of the method str.lstrip. Returns a list.
+
+    >>> lstrip(['Foo', 'Bar', 'Bam'], 'Foo')
+    ['Bar', 'Bam']
+
+    """
+    return list(lstrip_iter(iterable, strip_value))
```
mahmoud, line 190: One reason I was hoping to get @kurtbrose in here was to ask whether he would expect lstrip to dynamically handle the string case. I.e., autojoin things back together as "abc".lstrip('a') returns a string, not a list of `['b', 'c']`.

boltons/iterutils.py
```diff
@@ -177,6 +178,100 @@ def split_iter(src, sep=None, maxsplit=None):
...
+def strip_iter(iterable,strip_value=None):
+    """Strips values from the beginning and end of an iterable. Stripped items
+    will match the value of the argument strip_value. Functionality is 
+    analigous to that of the method str.strip. Returns a generator.
+
+    >>> list(strip_iter(['Fu', 'Foo', 'Bar', 'Bam', 'Fu'], 'Fu'))
+    ['Foo', 'Bar', 'Bam']
+
+    """
+    return rstrip_iter(lstrip_iter(iterable,strip_value),strip_value)
```
mahmoud, line 272: lstrip_iter and rstrip_iter I get, but just plain strip_iter? besides for consistency sake, how will this perform any differently than plain strip()?

boltons/iterutils.py
```diff
@@ -18,6 +18,7 @@
 import socket
 import hashlib
 import itertools
+import sys
```
mahmoud, line 21: is this import used?

tests/test_iterutils.py
```diff
@@ -509,3 +509,21 @@ def test_chunked_bytes():
     from boltons.iterutils import chunked
 
     assert chunked(b'123', 2) in (['12', '3'], [b'12', b'3'])
+
+
+def test_lstrip():
+    from boltons.iterutils import lstrip
+
+    assert lstrip([0,1,0,2,0,3,0],0) == [1,0,2,0,3,0]
```
mahmoud, line 517: should we test some more edge cases on these?

boltons/iterutils.py
mblahay: I wondered about that too but decided against that because it would be a special case to be used for strings only. These changes were meant to generically handle iterables, a class of value that strings happen to be part of. Besides, if this version of the function is problematic for a developer who is working with strings, there are always the string based methods.

boltons/iterutils.py
mblahay: I added strip because the standard String class includes lstrip, rstrip, and strip. I reasoned that since the "strip" functionality was to be emulated then all three needed to be available. Of course in reality strip as I coded it is it just joining lstrip and rstrip together to strip from both ends of the iterator.

As for your question about how strip_iter performs against "plain strip()." Not sure I understand. As was the pattern, all of the _iter functions are generators, while the non _iter functions return actual lists. From the standpoint of performance, if you are working with data of unknown size or very large amounts of data, you probably want to use the generators/iterators.

mblahay:
@mahmoud , is this ready for merging?

mahmoud:
@mblahay I think the short answer is: yes :) I'm cutting a release tonight and I figured I'd make the call. Thanks again for all your work and patience!

PR #259, opened 2020-05-25
fix(site): intersphinx link to recent PY2.7-->PY3

ankostis:
Update link used by `sphinx.ext.intersphinx` extension
to standard python docs, from PY2.7 --> PY alone (currently meant PY3),
so links to standard lib do not point to (almost) EOL Python.


mahmoud:
Been meaning to do this for a while, so a big thank you!

PR #262, opened 2020-07-27
Document that shutil.copytree now accepts dirs_exist_ok

jab:
boltons' copytree is needed for Python < 3.8.

Ref: python/cpython#8792


mahmoud:
Cool, thanks!

jab:
Thanks for merging!

PR #268, opened 2020-10-31
Bring Python current in classifiers

gliptak:


gliptak:
@mahmoud please review

mahmoud:
Another fine update, thanks!

PR #272, opened 2020-12-08
Bring Python versions current in Tox

gliptak:


gliptak:
@mahmoud please review

PR #276, opened 2021-01-19
Add license statement at start of every module

Jesse-Bakker:
The README installation section suggests vendoring some modules with applications. If care is not taken, it is easy to forget to include the license when doing so. This PR adds the license to the start of every module so that it is automatically included when copying files to ones own applications.

Feel free to close this if this is not a concern you share or you feel that this is not a good solution.

### DIFF
--- a/boltons/cacheutils.py
+++ b/boltons/cacheutils.py
@@ -1,4 +1,35 @@
 # -*- coding: utf-8 -*-
+
+# Copyright (c) 2013, Mahmoud Hashemi
+#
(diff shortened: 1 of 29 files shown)


mahmoud:
To be honest, as a concern it's pretty far from the top of my list. My top concern is anything that might prevent someone from saving time with a util that gets the job done, hence the vendorability. With this PR, I have a slight concern that the scary legal text might scare someone off, but because you've taken the time, and it's technically correct, I say let's give it a go. Thanks!

PR #277, opened 2021-02-10
Use constructor argument in `DeferredValue`

mallamanis:


mahmoud:
Just catching up here, good eye, thank you for catching this!

PR #278, opened 2021-02-20
Adds noop function satisfying PEP 559

CameronCairns:
https://www.python.org/dev/peps/pep-0559/

I've found recently a need to define a noop function when building functions that have logic that may need to be ran before or after a target function.  This is often used to setup some state necessary for the main function being ran or to run some cleanup functionality after a function is ran.  I found pep 559 when searching if python had a built in noop function, and saw that it was rejected.  I don't really want to have to define a noop function every time I work on a project that could use some noop functionality so I thought this would be perfect for boltons!

I thought the best location for this function is funcutils but please let me know if it isn't appropriate. Thanks for considering the PR!

### DIFF
--- a/boltons/funcutils.py
+++ b/boltons/funcutils.py
@@ -1075,4 +1075,26 @@ except ImportError:
                 setattr(cls, opname, opfunc)
         return cls
 
+def noop(*args, **kwargs):
+    """
+    Simple function that should be used when no effect is desired.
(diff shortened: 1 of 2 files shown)


mahmoud:
Makes total sense to me. Code looks good. CI's buggin (pip 21 failing to install under py2.7) but py3 is passing. Thanks!

PR #279, opened 2021-02-22
suport lists as key for bucketize

matiasmorant:
I was trying to do this and I naturally reached for `bucketize`, but then found it doesn't support lists

### DIFF
--- a/boltons/iterutils.py
+++ b/boltons/iterutils.py
@@ -575,11 +575,17 @@ def bucketize(src, key=bool, value_transform=None, key_filter=None):
     >>> bucketize(range(5), is_odd)
     {False: [0, 2, 4], True: [1, 3]}
 
-    *key* is :class:`bool` by default, but can either be a callable or a string
-    name of the attribute on which to bucketize objects.
+    *key* is :class:`bool` by default, but can either be a callable or a string or a list
+    if it is a string, it is the name of the attribute on which to bucketize objects.
 
     >>> bucketize([1+1j, 2+2j, 1, 2], key='real')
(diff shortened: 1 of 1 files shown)


mahmoud:
Interesting idea. I'm not against it, but I'm curious what happens if a value doesn't map to any of the specified buckets?

matiasmorant:
@mahmoud I don't understand the question. You have to specify the bucket for each value, so all values will map to a bucket (`src` and `key` are the same length)

kurtbrose:
huh, I guess you could achieve the same with:

```python
keys = [0, 1, 0]
bucketize('abc', lambda v: keys.pop())
{0: ['a', 'c'], 1: ['b']}
```

although that assumes the key function will be called once per value and in-order
(which is not a terrible assumption for a function which is meant to work with any iterable)

https://pypi.org/project/relativity/ could be helpful

```python
M2M(zip([0, 1, 0], 'abc'))
```

then you'd have, effectively, a dict-of-frozenset rather than dict-of-list

matiasmorant:
Sure, I could also do:
```
bucketize(zip(keys, values), key=lambda x: x[0], value_transform=lambda x: x[1]) 
```
but the point was to make it straightforward:
```
bucketize(values, keys) 
```

mahmoud:
Ah, got it. Let's raise a ValueError if the lengths of the lists aren't equal in that case.

FWIW I usually do a `zip` with an `dictutils.OMD().to_dict()`, but this is definitely more efficient if you don't need the rest of the OMD methods :)

matiasmorant:
@mahmoud  added the ValueError

mahmoud:
Looks good, thanks!

Issue #249, opened 2020-04-06
Add support for text mode in boltons.fileutils.atomic_save


mykter:
To allow atomic_save to be used as a straight-foward replacement for `open`, it would be nice if it supported a text mode, akin to `open(path, "w")`.
#131 describes a workaround for some situations, but it means atomic_save can't be a drop-in replacement. For example, it can't be easily used with `json.dump`, for which:

> The json module always produces str objects, not bytes objects. Therefore, fp.write() must support str input.

mahmoud:
Ah, it's actually there but not exposed in the docs! Try passing `text_mode=True` and see if that works for you. I'll get it added to the docs asap.

mahmoud:
OK added to the docs now. Good catch, keep an eye out! Thanks!

PR #281, opened 2021-03-06
Clarify docstring example for atomic_save

kylepjohnson:
This PR clarifies that `text_mode=True` means `w` and `text_mode=False` is `wb`. It also gives an example usage.

At first glance, I was confused by the docs for `atomic_save()` until duckduckgo lead me to issue #249 . If you don't want this, I won't be offended.

Also, [my project](https://github.com/cltk/cltk) uses `cachedproperty` extensively. Thank you, I don't know what we'd do without it!

### DIFF
--- a/boltons/fileutils.py
+++ b/boltons/fileutils.py
@@ -223,8 +223,12 @@ else:
 
 
 def atomic_save(dest_path, **kwargs):
-    """A convenient interface to the :class:`AtomicSaver` type. See the
-    :class:`AtomicSaver` documentation for details.
+    """A convenient interface to the :class:`AtomicSaver` type. Example:
+    
+    >>> with atomic_save("/tmp/file.txt", text_mode=True) as fo:
+    ...     data = fo.read()
+    
+    See the :class:`AtomicSaver` documentation for details.
     """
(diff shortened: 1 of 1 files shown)


mahmoud:
Hi Kyle! Far from not wanting it, this is fantastic. I think that's a great clarification. One note though: `text_mode` actually defaults to `False`, not `True`. Also, ignore the red CI, I've got to fix that :)

kylepjohnson:
Mahmoud, glad to be of service. Anything else I need to do for this PR?

boltons/fileutils.py
```diff
@@ -320,7 +324,7 @@ class AtomicSaver(object):
             respect the user's configured `umask`_, usually resulting
             in octal 0644 or 0664.
         text_mode (bool): Whether to open the destination file in text
-            mode.
+            mode (i.e., ``'w'`` not ``'wb'``). Defaults to ``True``.
```
mahmoud, line 327:
```suggestion
            mode (i.e., ``'w'`` not ``'wb'``). Defaults to ``False`` (``wb``).
```

mahmoud:
All good, I committed the suggestion. Merged, thanks again!

PR #283, opened 2021-03-19
Update Python 3.9 FrozenDict exception message

bnavigator:
The exception message with Python 3.9.2 is 'FrozenDict object is immutable'


mahmoud:
Looks about right to me, thank you so much!

PR #286, opened 2021-05-09
Tox GH Action

mahmoud:


mahmoud:
Just dug out an old windows machine to confirm that yes, GitHub Windows runners are correct, and AppVeyor Windows runners have been wrong this whole time. That said, TextIOWrapper newline handling mixes particularly poorly with multicharacter newlines and block boundaries.

PR #159, opened 2018-05-31
Add test cases for issue #158

wcooley:
[pytest-out.txt](https://github.com/mahmoud/boltons/files/2060236/pytest-out.txt)

PR #298, opened 2022-01-09
Chaining `boltons.urlutils.URL.navigate()` results in `TypeError`

### LINKED ISSUE #158: URL.navigate fails with path in base URL

allanlei:
Chaining `navigate()` calls like `boltons.urlutils.URL('...').navigate(...).navigate(...)` produces `TypeError: can only concatenate list (not "tuple") to list`.

`boltons.urlutils.URL('...').navigate(...).navigate(...)` should be the equivalent of 
```python
url = boltons.urlutils.URL('...').navigate(...)
url = boltons.urlutils.URL(url).navigate(...)
```

This PR extends #159 as the test cases provided are incosistent if it was manually chained.

Fixes #158, #159, #160

### DIFF
--- a/boltons/urlutils.py
+++ b/boltons/urlutils.py
@@ -654,7 +654,8 @@ class URL(object):
             if dest.path.startswith(u'/'):   # absolute path
                 new_path_parts = list(dest.path_parts)
             else:  # relative path
-                new_path_parts = self.path_parts[:-1] + dest.path_parts
+                new_path_parts = list(self.path_parts[:-1]) \
+                               + list(dest.path_parts)
         else:
             new_path_parts = list(self.path_parts)
             if not query_params:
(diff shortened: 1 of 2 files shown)


mahmoud:
Nice work, appreciate the tests! Also check out [hyperlink](https://github.com/python-hyper/hyperlink). It's a bit different, but might fit the bill in some cases.  :) Thanks again!

Issue #290, opened 2021-07-20
__version__ field


kasium:
Hey,

thanks for this awesome library. Can you please consider to add a \_\_version__ field to boltons, so that something like the following is possible?

```python
import boltons

print(boltons.__version__)
# 21.1.1
```

Thanks,
Kai

mahmoud:
Hey Kai! Thanks for the kind words!

boltons started out very piecemeal, with each module designed to be standalone, which is why this doesn't already exist. Of course, that's not a reason why it shouldn't exist. So I'll try to make this happen next time I'm in the neighborhood, assuming someone doesn't beat me to it :)

kasium:
@mahmoud if it would be fine for you, I can also give it a try

PR #300, opened 2022-02-27
Update package metadata

ofek:
## Background

Hello there! The Python packaging ecosystem has now standardized on the interface for build backends ([PEP 517](https://www.python.org/dev/peps/pep-0517/)/[PEP 660](https://www.python.org/dev/peps/pep-0660/)) and the format for metadata declaration ([PEP 621](https://www.python.org/dev/peps/pep-0621/)/[PEP 631](https://www.python.org/dev/peps/pep-0631/)).

As such, I'm spending my free time updating important projects so that they are modernized and set an example for others 😄 

## Summary of changes

This implements [PEP 621](https://www.python.org/dev/peps/pep-0621/), obviating the need for `setup.py`, `setup.cfg`, and `MANIFEST.in`. Support has not landed in `setuptools`, so builds will now use `hatchling`.

I've done this for such projects as [pipx](https://github.com/pypa/pipx/blob/2ecc668acf472ad6956cc682499c077c1130d17e/pyproject.toml), all [Datadog Agent integrations](https://github.com/DataDog/integrations-core/pulls?q=is%3Apr+author%3Aofek+in%3Atitle+Add+pyproject.toml+file), etc.

## Notes

- This also drops support for EOL Python 3.x versions since `hatchling` and other build backends like `setuptools` have already done so. This will not break anything because `pip` only looks for version candidates that satisfy the `requires-python` field. Here are stats for the curious: https://pypistats.org/packages/boltons
- This also resolves https://github.com/mahmoud/boltons/issues/290 as I saw that had not yet been handled, @mahmoud let me know if that's okay!

### DIFF
--- a/boltons/__init__.py
+++ b/boltons/__init__.py
@@ -1,1 +1,1 @@
-__version__ = '21.0.1dev'
+__version__ = '22.0.0dev'


boltons/__init__.py
```diff
@@ -0,0 +1 @@
+__version__ = '21.0.1dev'
```
mahmoud, line 1:
```suggestion
__version__ = '22.0.0dev'
```

mahmoud (changes requested):
Wow! Rocketing into the future here. boltons is a simple library, so this makes a lot of sense, and I appreciate you taking the time! 

One small change, and a couple thoughts/concerns:

1. Hatchling seems very new (still [0ver](https://0ver.org/) ;) ). What's the plan for maintenance/stability?
2. Does Hatch/Hatchling offer anything like packaging tests we can add to the tox.ini?

Thanks in advance!

ofek:
> Hatchling seems very new (still [0ver](https://0ver.org/) ;) ). What's the plan for maintenance/stability?

It is quite stable and actively maintained, the only reason the version is 0.x is because 1.0.0 will drop support for Python 2.

> Does Hatch/Hatchling offer anything like packaging tests we can add to the tox.ini?

If you mean like checking the contents of the sdist `*.tar.gz` file then no, I don't think any of the newer backends do that because there is no more `MANIFEST.in` which was the confusing and error prone part.

ofek:
Oh btw it's also on [conda-forge](https://anaconda.org/conda-forge/hatchling), [Debian](https://salsa.debian.org/python-team/packages/hatchling), [Fedora](https://src.fedoraproject.org/rpms/python-hatchling), etc.

ofek:
@mahmoud Would you mind triggering the CI with that button below?

mahmoud:
> Would` you mind triggering the CI with that button below?
Done :) All green.

>> Hatchling seems very new (still [0ver](https://0ver.org/) ;) ). What's the plan for maintenance/stability?
> It is quite stable and actively maintained, the only reason the version is 0.x is because 1.0.0 will drop support for Python 2.

I guess my main concern is that Hatchling first came out (0.3) two months ago? And I love the velocity, but it also makes me nervous should something break. Presumably I'd own the fix at that point :) 

Longer term, what's your recommendation when setuptools support lands?

> If you mean like checking the contents of the sdist *.tar.gz file then no, I don't think any of the newer backends do that because there is no more MANIFEST.in which was the confusing and error prone part.

MANIFEST.in could certainly be tricky, but even in this modern approach, couldn't [the includes here](https://github.com/mahmoud/boltons/pull/300/files#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711R45) go out of sync if a new file is added?

Finally, maybe a very simple question, but if someone were to seek to install boltons from a source distribution built with this modern approach, how would it be recommended they do that without a setup.py? Does pip know what to do these days?

ofek:
> should something break. Presumably I'd own the fix at that point

No it would be on the maintainers, just as if you used `setuptools` or some other build backend. Since we use it at Datadog for shipping business critical things, our CI would be the first to yell at me 😄 

-----

> Longer term, what's your recommendation when setuptools support lands?

I still would not use it. I can briefly tell you why we moved:

Our goal is to remove all `setup.py` files since their use is now [heavily discouraged](https://blog.ganssle.io/articles/2021/10/setup-py-deprecated.html) and build with a [PEP 517](https://www.python.org/dev/peps/pep-0517/) compatible build system while defining metadata with [PEP 621](https://www.python.org/dev/peps/pep-0621/).

### setuptools

Disadvantages:

- Does not yet support `pyproject.toml`-based metadata declaration nor editable installation standards ([PEP 660](https://www.python.org/dev/peps/pep-0660/))
- Based on a [deprecated](https://www.python.org/dev/peps/pep-0632/) standard library module that it now vendors
- Slow to get things merged since the code quality issue is exacerbated by the fact that it is mostly maintained by just [one person](https://github.com/jaraco)

### Flit

[`flit-core`](https://pypi.org/project/flit-core/) is the build backend behind [Flit](https://github.com/pypa/flit).

Disadvantages:

- No mechanism for extending behavior by design, so we could never do things like versioning with Git tags
- Very opinionated about project layouts since it targets beginners e.g. [only supports](https://github.com/pypa/flit/issues/370#issuecomment-973359998) implicit namespace packages, must define certain data like `__version__` inside the root `__init__.py` file, etc.

### Poetry

[`poetry-core`](https://pypi.org/project/poetry-core/) is the build backend behind [Poetry](https://github.com/python-poetry/poetry).

This was actually a no-go option for us due to its persistent unwillingness to adopt new standards if they are deemed suboptimal (see comments on [PEP 621](https://discuss.python.org/t/5243/12) and [PEP 665](https://github.com/python-poetry/poetry/issues/4710#issuecomment-973946104))

### Hatchling

- In comparison to `setuptools`, Hatchling is far smaller, making it less susceptible to security issues and more auditable. For example, the logic to build a wheel resides in this [one file](https://github.com/ofek/hatch/blob/master/backend/src/hatchling/builders/wheel.py).
- In comparison to `flit-core`, Hatchling is extensible by design. For example, there is already [a plugin](https://github.com/ofek/hatch-vcs) for [setuptools_scm](https://github.com/pypa/setuptools_scm) (which despite the legacy naming is actually now decoupled from `setuptools`) and [a plugin](https://github.com/ofek/hatch-mypyc) for compiling code with [Mypyc](https://github.com/mypyc/mypyc).
- In comparison to `poetry-core`, Hatchling never strays from community standards and is in fact eager to adopt new ones (it's the only backend that currently supports [PEP 639](https://www.python.org/dev/peps/pep-0639/)).

-----

> couldn't [the includes here](https://github.com/mahmoud/boltons/pull/300/files#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711R45) go out of sync if a new file is added

Well yes of course. If you want everything added to the source distribution that section can just be omitted, or define [exclude](https://ofek.dev/hatch/latest/config/build/#patterns).

-----

> if someone were to seek to install boltons from a source distribution built with this modern approach, how would it be recommended they do that without a setup.py? Does pip know what to do these days?

Yes, since version [19.0](https://pip.pypa.io/en/stable/news/#v19-0).

ofek:
Just let me know if you want me to do anything else here!

bwatson-dk:
You could also just use the setuptools build backend.

```yaml
build-backend = "setuptools.build_meta"
```

It's pretty standardized and I don't think there are any concerns about maintenance.

ofek:
> It's actually possible to have a pyproject.toml as the only configuration file.

Support landed a few days ago and is experimental. See notes and fixes: https://setuptools.pypa.io/en/latest/history.html

> I don't think there are any concerns about maintenance

True but in comparison to `setuptools`, Hatchling has reproducible builds, supports the modern [PEP 660](https://peps.python.org/pep-0660/) editable (`pip install -e ...`) installation standard, has a more intuitive plugin system, and the codebase is an order of magnitude smaller thus making it more auditable and less prone to bugs (`setuptools` is based on a [deprecated](https://peps.python.org/pep-0632/) standard library module that it now must vendor).

induane:
I'm not sure having ``distutils`` in ``setuptools`` is a huge issue, it was actually removed from Python in part BECAUSE it was already in ``setuptools`` anyway and there wasn't a reason to maintain it in the standard library. **Full** support for ``pyproject.toml`` WAS only recently merged, but it's been possible to include metadata there for a while but it didn't validate it and defaulted to ``setup.cfg`` in any conflict; I've been doing so for a spell and with the new release the whole thing is working even better.

Really though @ofek I'm not disagreeing with your analysis; my swing through Hatchling actually impressed me and I'm going to take it for a spin on a personal project so I have a reminder to keep an eye on it. It may be  (probably even IS?) the superior project. I hope it takes off!

Really though, technical analysis aside, my preference for widespread important projects like ``boltons`` tends to be conservative. For better or worse ``setuptools`` is the major player here. It's not the only one but it's the old hat in the ring and it's the conservative choice. Were Hatchling to really catch on and take off I'd say it would be a really cool thing to look at upgrading to. Swapping build backends isn't too hard anyway.

Of course it's not my call at all :smile:  - I was just popping by to leave an unsolicited opinion :wink:

ofek:
Small update: Hatch will likely be adopted by PyPA in the next week

- https://mail.python.org/archives/list/pypa-committers@python.org/thread/J2BL5XTTOCXUCELUE7L5P5HHJ7NXZGZ5/
- https://mail.python.org/archives/list/pypa-committers@python.org/thread/ZWSSNKQJ3GRIAK73FJS6KHOJNUJZYOVX/

Nothing major will change, just the repo location

ofek:
Transfer complete! https://github.com/pypa/hatch

PR #302, opened 2022-03-15
Fix infinite daterange issue when start and stop is identical

### LINKED ISSUE #297: daterange(start, stop, inclusive=True) returns an infinite iterator when start is same as stop
expected to get `start` exactly once.

mehq:
Fixes #297.


mahmoud:
Good catch, and good fix. The test is much appreciated, as well. Thanks for this!

PR #303, opened 2022-03-16
Fix implicit relative imports

skravik:
Triggers caught ImportError fallbacks when run on Python 3, which can lead to subtle unexpected behavior. Some files already had the correct syntax, this PR updates the remainder.

### DIFF
--- a/boltons/cacheutils.py
+++ b/boltons/cacheutils.py
@@ -82,7 +82,7 @@ except Exception:
             pass
 
 try:
-    from boltons.typeutils import make_sentinel
+    from .typeutils import make_sentinel
     _MISSING = make_sentinel(var_name='_MISSING')
     _KWARG_MARK = make_sentinel(var_name='_KWARG_MARK')
 except ImportError:
--- a/boltons/debugutils.py
+++ b/boltons/debugutils.py
@@ -47,7 +47,7 @@ except NameError:
     from reprlib import Repr
 
 try:
-    from typeutils import make_sentinel
+    from .typeutils import make_sentinel
     _UNSET = make_sentinel(var_name='_UNSET')
 except ImportError:
     _UNSET = object()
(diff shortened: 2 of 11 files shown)


bwatson-dk:
Part of the way boltons is structured is to make it easy to partially vendor it. For instance, using ``debugutils.py`` should be as simple as dropping ``debugutils.py`` and ``typeutils.py`` into any folder. The absolute import would break this (though only trivially). Maybe they should be:

```python
from .typeutils import make_sentinal
```

Instead?

skravik:
Good points, ease of vendoring may be a reason to keep the (explicit) relative imports. The PR was just aiming to maintain consistency with the few existing places in the code base that were already using an absolute style. Could certainly update those too though:

https://github.com/mahmoud/boltons/blob/395f690f4a24331c4554e2169ac18a15955a4eab/boltons/cacheutils.py#L85
https://github.com/mahmoud/boltons/blob/395f690f4a24331c4554e2169ac18a15955a4eab/boltons/funcutils.py#L67
https://github.com/mahmoud/boltons/blob/395f690f4a24331c4554e2169ac18a15955a4eab/boltons/urlutils.py#L1541

I don't have a strong preference, so whichever the author prefers. I'm happy to update PR as needed.

mahmoud:
Hey! Thanks for this (and the patience in awaiting reply).  As I'm sure you've gathered, these sentinels are best effort, and mostly for repr/docs generation niceness (hence the ImportError catching for folks who vendor). 

But, we don't want to fill logs with warnings if we don't have to. I think I do prefer the explicit relative imports, assuming it passes all tests. Could you rebase and make that change? Thanks!

skravik:
Rebased and updated. Thanks!

mahmoud:
LGTM, thanks again!

PR #305, opened 2022-03-29
Fixes and improvements to SpooledBytesIO & SpooledStringIO

bwatson-dk:
- Inherit from ``io.IOBase`` so instance & type checks are simpler
- ``__eq__`` and ``__ne__`` no longer need to load both files completely into memory
- ``__eq__`` and ``__ne__`` reset read-head in comparison errors
- Some methods now raise a ValueError if the file is closed. This brings the implementations closer to ``StringIO`` and implementations from the ``io`` module
- Added some additional unittests
- Removed some Python 2.6 style format strings

### DIFF
--- a/boltons/ioutils.py
+++ b/boltons/ioutils.py
@@ -40,7 +40,7 @@ are useful when dealing with input, output, and bytestreams in a variety of
 ways.
 """
 import os
-from io import BytesIO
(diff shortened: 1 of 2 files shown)


mahmoud:
Hey, these updates look great! Thanks for this!

PR #307, opened 2022-04-10
Added py310 to gh actions

suimong:
- Updated packages in requirements-test.txt to latest versions: the current version is incompatible with python 3.10
- Added a separate requirements-test-legacy.txt for py27, py34, py36, pypy, pypy3

### DIFF
--- a/requirements-test.txt
+++ b/requirements-test.txt
@@ -1,4 +1,4 @@
-coverage==4.5.1
-pytest==3.0.7
-pytest-cov==2.5.1
-tox<3.0
+coverage==6.2
+pytest==7.1.1
+pytest-cov==3.0.0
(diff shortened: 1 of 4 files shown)


requirements-test.txt
```diff
@@ -1,4 +1,4 @@
-coverage==4.5.1
-pytest==3.0.7
-pytest-cov==2.5.1
-tox<3.0
+coverage==6.2
+pytest==7.1.1
+pytest-cov==3.0.0
+tox==3.24.5
```
guyarad, line 4: maybe we should less restrictive versioning?

suimong: Definitely. Restricted to major version number in new commit.

Issue #310, opened 2022-05-26
windowed with overlap


wimglenn:
Hello, thank you for this super useful and high-quality library! 
I'm wondering if there is any support for [windowed](https://boltons.readthedocs.io/en/latest/iterutils.html#boltons.iterutils.windowed_iter) with overlap, e.g. by default there is overlap of the window size - 1

```
>>> g = windowed_iter(itertools.count(), 3)
>>> next(g)
(0, 1, 2)
>>> next(g)
(1, 2, 3)
>>> next(g)
(2, 3, 4)
```

But what if you want (0, 1, 2), (2, 3, 4), (4, 5, 7)... ?  Any easy solution in boltons?

PR #312, opened 2022-06-24
add chunk_ranges function to iterutils

jstriebel:
This PR adds a `chunk_ranges` function to iterutils, which doesn't chunk an iterable or list directly, but produces the relevant indices to chunk it. It contains functionality for windowing/overlap (see also issue https://github.com/mahmoud/boltons/issues/310) and alignment of the chunks. I'm not sure if `input_size` or `input_end` is more appropriate for the arguments, also the function name might need improvement.

It would be great to incorporate this function into boltons IMO, thanks a lot for considering this.

This PR includes the function itself, a test, and updates the docs.

### DIFF
--- a/boltons/iterutils.py
+++ b/boltons/iterutils.py
@@ -323,6 +323,13 @@ def chunked(src, size, count=None, **kw):
         return list(itertools.islice(chunk_iter, count))
 
 
+def _validate_positive_int(value, name, strictly_positive=True):
+    value = int(value)
+    if value < 0 or (strictly_positive and value == 0):
+        raise ValueError('expected a positive integer ' + name)
+    return value
+
+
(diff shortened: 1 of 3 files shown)


jstriebel:
@mahmoud Friendly ping :-) What do you think about adding those changes?

mahmoud (approved):
Hey thanks for your patience (#newdadlife over here), and thanks for the PR! 

From what I see:

* It's performing non-trivial work
* It's tested
* It's documented
* It follows the [boltons architecture](https://boltons.readthedocs.io/en/latest/architecture.html)

So let's go for it :)

jstriebel:
Thank *you* for maintaining this lib, @mahmoud!

PR #315, opened 2022-08-18
bugfix in mathutils.Bits.as_list

kurtbrose:
list(Bits(0, 2)) == [False, False] -- which is the correct behavior
Bits(0, 2).as_list() == [False] -- which was a bug


boltons/mathutils.py
```diff
@@ -215,7 +215,7 @@ def __hash__(self):
         return hash(self.val)
 
     def as_list(self):
-        return [c == '1' for c in '{0:b}'.format(self.val)]
+        return [c == '1' for c in  '{{0:0{0}b}}'.format(self.len).format(self.val)]
```
nichmor, line 218: I see that below we have a function named as_bin
what do you think about rewriting it like this?
`return [c == '1' for c in  self.as_bin()]` ? 
Also, what do you think about adding a test for this case?

PR #318, opened 2022-10-15
types boltons/dictutils.py

blablatdinov:


boltons/dictutils.pyi
```diff
@@ -0,0 +1,118 @@
+try:
+    from collections.abc import ItemsView, KeysView, ValuesView
+except ImportError:
+    from collections import KeysView, ValuesView, ItemsView
```
nichmor, line 4:
```suggestion
    from collections import ItemsView, KeysView, ValuesView
```
in my opinion we should list them in the same order, for consistency

boltons/dictutils.pyi
```diff
@@ -0,0 +1,118 @@
...
+    def keys(self, multi: bool = ...) -> list[Hashable]: ...
+    def values(self, multi: bool = ...) -> list[Any]: ...
+    def items(self, multi: bool = ...) -> list[Generator[Any, None, None]]: ...
+    def __iter__(self) -> Generator[Any, None, None]: ...
+    def __reversed__(self) -> Generator[Any, None, None]: ...
+    def viewkeys(self) -> KeysView: ...
+    def viewvalues(self) -> ValuesView: ...
+    def viewitems(self) -> ItemsView: ...
+
+OMD = OrderedMultiDict
```
nichmor, line 55: Why do we need it here? it is not used anywhere

mahmoud: I think these aliases exist in the module? Doesn't seem like they'll do any harm...

mahmoud (commented):
Very interesting! boltons's first .pyi, thank you for this. I have two high-level questions:

1. I don't believe these are enforced anywhere. Is there a pytest or tox command that you'd recommend to validate these are correct?
1. Any strong reasons for `.pyi` instead of embedding the typehints?

Thanks so much!

Avasam:
Now that types can be used inline (Python 3.7 still has *some* challenges), checkout https://github.com/mahmoud/boltons/pull/382

PR #332, opened 2023-03-30
Update tbutils.py to parse tracebacks with anchor lines

sestinj:
Anchors are sometimes added below the source line, (see https://github.com/python/cpython/blob/dcd6f226d6596b25b6f4004058a67acabe012120/Lib/traceback.py#L479-L510). This fix will allow parsing to continue as normal in their presence

### DIFF
--- a/boltons/tbutils.py
+++ b/boltons/tbutils.py
@@ -682,7 +682,7 @@ def fix_print_exception():
 _frame_re = re.compile(r'^File "(?P<filepath>.+)", line (?P<lineno>\d+)'
                        r', in (?P<funcname>.+)$')
 _se_frame_re = re.compile(r'^File "(?P<filepath>.+)", line (?P<lineno>\d+)')
-
+_underline_re = re.compile(r'^[~^ ]*$')
 
 # TODO: ParsedException generator over large bodies of text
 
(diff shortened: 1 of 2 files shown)


mahmoud:
Ohhh very very nice, yes I was just running into something related today. Any chance we could get a test over here? https://github.com/mahmoud/boltons/blob/master/tests/test_tbutils_parsed_exc.py

Thanks for getting the ball rolling in either case!

sestinj:
Of course, here's a test! Added a note, but obviously without storing more information about columns in ParsedException (already noted [here](https://github.com/mahmoud/boltons/blob/f892a2b46e7fd31c639723d1d9bf4d9ad63752d6/boltons/tbutils.py#L694-L696) that some of this is missing), this will break the `to_string(from_string(original_tb)) == original_tb` property.

mahmoud:
Good point. I'm not opposed to adding more attributes, of course. Any interest in pursuing that in this PR, or should we make a separate issue?

sestinj:
It's unlikely I'll get to that soon, so wouldn't want to hold up this PR. I made a new issue [here](https://github.com/mahmoud/boltons/issues/333).

mahmoud:
Sounds good, thanks again for this!

PR #336, opened 2023-04-12
fix: make copy_function work with keyword-only defaults

thorwhalen:
At this point, we have:

```python
from boltons.funcutils import copy_function

f = lambda x, *, y=2: x * y

f_copy = copy_function(f)

assert f(21) == 42  # all goes well, but...

import pytest, re

with pytest.raises(TypeError):
    f_copy(21)  # missing 1 required keyword-only argument: 'y'
    
```

So `f_copy` isn't really a copy.

By adding

```python
...
    if hasattr(orig, "__kwdefaults__"):
        ret.__kwdefaults__ = orig.__kwdefaults__
...
```

to the function, we get the desired behavior:

```python
f = lambda x, *, y=2: x * y
f_copy = copy_function(f)
assert f(21) == f_copy(21) == 42 
```

I also extended the `test_copy_function` test function to protect this behavior in future evolutions.

### DIFF
--- a/boltons/funcutils.py
+++ b/boltons/funcutils.py
@@ -215,6 +215,8 @@ def copy_function(orig, copy_dict=True):
                        name=orig.__name__,
                        argdefs=getattr(orig, "__defaults__", None),
                        closure=getattr(orig, "__closure__", None))
+    if hasattr(orig, "__kwdefaults__"):
+        ret.__kwdefaults__ = orig.__kwdefaults__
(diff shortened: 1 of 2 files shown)


mahmoud:
This is great Thor, thanks for the great report and succinct test + fix. I moved the test into the py3-only funcutils tests and kicked off discussions for dropping py2 support via #339. Will get this out in the next release, hopefully in the next week or so.

PR #338, opened 2023-04-18
Make OMD picklable under py3, fixes #337

### LINKED ISSUE #337: Non-empty `dictutils.OMD` cannot be loaded from `pickle`
A non-empty `OrderedMultiDict` can be serialized, but cannot be deserialized.
A empty `OrderedMultiDict` can be both serialized and deserialized.

* Python `3.11.3`
* Version `23.0.0`

```python
In [1]: import pickle

In [2]: import boltons.dictutils

In [3]: pickle.loads(pickle.dumps(boltons.dictutils.OMD()))
Out[3]: OrderedMultiDict([])

In [4]: pickle.loads(pickle.dumps(boltons.dictutils.OMD({'a': 1})))
AttributeError                            Traceback (most recent call last)
Cell In[4], line 1
----> 1 pickle.loads(pickle.dumps(boltons.dictutils.OMD({'a': 1})))

File /usr/local/lib/python3.11/site-packages/boltons/dictutils.py:333, in OrderedMultiDict.__setitem__(self, k, v)
    331 if super(OrderedMultiDict, self).__contains__(k):
    332     self._remove_all(k)
--> 333 self._insert(k, v)
    334 super(OrderedMultiDict, self).__setitem__(k, [v])

File /usr/local/lib/python3.11/site-packages/boltons/dictutils.py:198, in OrderedMultiDict._insert(self, k, v)
    197 def _insert(self, k, v):
--> 198     root = self.root
    199     cells = self._map.setdefault(k, [])
    200     last = root[PREV]

AttributeError: 'OrderedMultiDict' object has no attribute 'root'
```

mahmoud:
For some reason this test passed under py2, which I guess is telling as to how long it's been since I was actively pickling.

### DIFF
--- a/boltons/dictutils.py
+++ b/boltons/dictutils.py
@@ -71,8 +71,10 @@ thanks to `Mark Williams`_ for all his help.
 
 try:
     from collections.abc import KeysView, ValuesView, ItemsView
+    PY3 = True
 except ImportError:
     from collections import KeysView, ValuesView, ItemsView
+    PY3 = False
 
 import itertools
 
@@ -177,7 +179,7 @@ class OrderedMultiDict(dict):
         ret = super(OrderedMultiDict, cls).__new__(cls)
         ret._clear_ll()
         return ret 
-
+    
     def __init__(self, *args, **kwargs):
         if len(args) > 1:
             raise TypeError('%s expected at most 1 argument, got %s'
@@ -188,6 +190,14 @@ class OrderedMultiDict(dict):
             self.update_extend(args[0])
         if kwargs:
             self.update(kwargs)
+
+    if PY3:
+        def __getstate__(self):
+            return list(self.iteritems(multi=True))
+        
+        def __setstate__(self, state):
+            self.clear()
+            self.update_extend(state)
 
     def _clear_ll(self):
         try:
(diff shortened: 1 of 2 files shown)


allanlei (changes requested):
There seems to be additional pickling issues.  The deserialized value loses the multi-value for `getlist`.

```python
In [48]: value = OrderedMultiDict([('a', 1), ('a', 2), ('b', 3)])

In [49]: value.getlist('a')
Out[49]: [1, 2]


In [50]: pickle.loads(pickle.dumps(value)).getlist('a')
Out[50]: [2]

# `iteritems` seems to be correct
In [16]: list(value.iteritems(multi=True))
Out[16]: [('a', 1), ('a', 2), ('b', 3)]

In [11]: list(pickle.loads(pickle.dumps(value)).iteritems(multi=True))
Out[11]: [('a', 1), ('a', 2), ('b', 3)]
```

mahmoud:
Right you are. Odd that Py2 was so much better behaved than Py3. 

Just a couple implementation notes:

- The `__getnewargs__` approach in the docs almost works, but because OMD inherits from `dict`, the builtin implementation overwrites the values via `__setitem__` calls after `__new__`. That's why only individual values made it through; OMD's `__setitem__` clears previously-added values.
- `__setstate__` apparently runs after the `__setitem__` calls mentioned above, allowing us to reset the multi-value. This does mean that there's some extra work happening during unpickling. Maybe a fancy `__reduce_ex__` could get around this, but the current approach seems fine for now.

Issue #344, opened 2023-07-04
chunked filter


rafalkrupinski:
Is boltons a good home for a filter_chunked function?

```python
def filter_chunked(iterable: Iterable[T], chunk_size: int, func: Callable[[Iterable[T]], Iterable[bool]]) -> Iterator[T]:
    for chunk in boltons.iterutils.chunked(iterable, chunk_size):
        for item, allow in zip(chunk, func(chunk)):
            if allow:
                yield item
```

It behaves similar to filter(), except the predicate is called with chunks of input. It's useful the predicate must make calls to external services, but can do it in batches.

mahmoud:
Hey Raphael! So, `filter_chunked` is certainly adjacent to some things I've done in the past. Just at first blush, I wondered how this might be rewritten as a composition of `chunked_iter`, `zip`, and `filter`. And it turned out to be pretty annoying!

(something like:
```python
from boltons.iterutils import chunked_iter

key = lambda chunk: [bool(x%2) for x in chunk]
vals = list(range(20))
chunk_size = 5

filtered_chunked = (x for chunk in chunked_iter(vals, chunk_size) for x, allow in zip(chunk, key(chunk)) if allow)

# list(filtered_chunked) -> [1, 3, 5, 7, 9, 11, 13, 15, 17, 19]
```

If it's this annoying, I'm generally for including it. What do you think of the simplified expression above? You'll definitely want to use `chunked_iter` if you're going to have make a generator. We may need to name it `filter_chunked_iter`, too.

The input iterable is flat, and the output iterable is flat, so that dimensionality matches, and there's no need for a `value_transform` convenience like with [`bucketize`](https://boltons.readthedocs.io/en/latest/iterutils.html#boltons.iterutils.bucketize). 

We'll also need a nice, demonstrative docstring. What's your motivating example in this case? Are there a lot of APIs where this batch nature applies? 

Thanks again!

rafalkrupinski:
> What do you think of the simplified expression above?

I'm happy with a generator expression, I'd just split it in lines and use a nicer variable name for my PR.
I wouldn't use `key` as the function-param name, bc it suggests it returns some inherent property of the object (like in `sorted`). I did use just `func` above but I think something `batch_pred` or `chunk_pred` (predicate) would suit.

> What's your motivating example in this case?

My code is looking for resources that exist in one API but not in another one. Both APIs use paging, have limits on query length, and allow for `property in $values` semantics.

[...]

PR #346, opened 2023-07-09
Add chunked_filter (#344)

rafalkrupinski:


boltons/iterutils.py
```diff
@@ -374,6 +374,30 @@ def chunked_iter(src, size, **kw):
     return
 
 
+def chunked_filter(iterable, predicate, chunk_size):
```
mahmoud, line 377: Great to see a docstring here. Looking around at other functions, I noticed that most refer to the predicate as `key`. Do you mind updating this argument to be consistent?

boltons/iterutils.py
```diff
@@ -374,6 +374,30 @@ def chunked_iter(src, size, **kw):
...
+    """A version of :func:`filter` which will call predicate with a chunk of the iterable.
```
mahmoud, line 378: I think this is a great description, though it's a good idea to style arguments when referring to them (italicize or monospace).

mahmoud (changes requested):
Overall looking good, would also be good to see an extra test or two for edge cases in `test_iterutils.py`. Thanks again for this!

boltons/iterutils.py
rafalkrupinski: I see various names. builtins.filter uses `iterable` and `function`. Actually it's [filter(function, iterable)](https://docs.python.org/3/library/functions.html#filter). Since it's a variation of it, how about using these names and order?

rafalkrupinski: interestingly [itertools.filterfalse(predicate, iterable)](https://docs.python.org/3/library/itertools.html#itertools.filterfalse)

rafalkrupinski:
builtins.filter allows None function, in which case an identity function is used. Here it doesn't make much sense from the use case perspective, but perhaps it would from the usability perspective and for consistency?

boltons/iterutils.py
```diff
@@ -374,26 +374,50 @@ def chunked_iter(src, size, **kw):
     return
 
 
-def chunked_filter(iterable, predicate, chunk_size):
-    """A version of :func:`filter` which will call predicate with a chunk of the iterable.
+def chunked_filter(iterable, predicate, size):
```
mahmoud, line 377: oops, based on the docstring, I think you meant to rename `predicate` here to `key`.

boltons/iterutils.py
```diff
@@ -374,26 +374,50 @@ def chunked_iter(src, size, **kw):
...
     [0, 2, 4, 6, 8]
 
     In the above example the lambda function is called twice: once with values
     0-4 and then for 5-9.
 
     Args:
         iterable (Iterable): Items to filter
-        predicate (Callable): Predicate function
-        chunk_size (int): The maximum size of chunks that will be passed the
+        predicate (Callable): Bulk predicate function that accepts a list of items
```
mahmoud, line 388: `predicate` -> `key` to match description above?

boltons/iterutils.py
```diff
@@ -374,26 +374,50 @@ def chunked_iter(src, size, **kw):
...
+        raise TypeError('expected callable key')
+
+    def predicate_(src_):
+        allow_iter = predicate(src_)
+        if not is_iterable(allow_iter):
+            raise TypeError('expected an iterable from key(src)')
+
+        allow_list = list(allow_iter)
+        if len(allow_list) != len(src_):
+            raise ValueError('expected the iterable from key(src) has the same length as the passed chunk of items')
```
mahmoud, line 413: "has the same" -> "to have the same"

I think it's great you're thinking about this. For exceptions, these days, I usually recommend the following format to maximize debuggability:

`raise ValueError("chunked_filter expected key func to return an iterable of length {src_len}, got {allow_list_len}")`.

Similar changes could be made to other exception messages above just by adding ", not {actually_received_type}"

mahmoud (changes requested):
Good progress! I like all the edge case handling. I agree that there's no harm in allowing `key` to be `None`, except that then you may have to come up with a default for `size`, based on the ordering. I'm fine either way, so I'll let you make the call. :)

boltons/iterutils.py
rafalkrupinski: not sure, see previous comments

boltons/iterutils.py
mahmoud: Ah, I think the prime reference for this should be the [`sorted()`](https://docs.python.org/3/library/functions.html#sorted) function, where `key` is used. `min()` and `max()` also use `key`, I think. So, tldr. `key` should be preferred here. :)

rafalkrupinski: > Ah, I think the prime reference for this should be the [`sorted()`](https://docs.python.org/3/library/functions.html#sorted) function, where `key` is used.

There's a good reason why sorted and min/max use `key` and `filter` & `filterfalse` don't. When you compare two objects, you either compare the value itself or its part, hence the `key` - attribute, only shorter. When you filter a collection, it can be an arbitrary criterium that's external to the object, so you use `function` (...to be called on an item) or more specifically, `predicate`.
Also, since it's an extension of `builtins.filter`, what's the reason for following naming pattern of any other function?

Issue #348, opened 2023-10-10
LRU .values() and dict return old entries


nicoddemus:
Hi,

First of all thanks for the excellent library!

I'm seeing strange results when using `LRU`: after replacing an existing entry with a new value, seems like the `keys()` and `values()` methods return the old value, instead of the new one. Using `__getitem__` and `.get` we have the expected behavior.

Here is a test showing this behavior:

```python
def test_lru():
    from boltons.cacheutils import LRU

    cache = LRU()

    # Add an entry.
    cache['a'] = 1
    
    # Normal __getitem__ access. 
    assert cache['a'] == 1  # passes.
    # Convert to dict.
    assert dict(cache) == {'a': 1}  # passes.
    # Another way to access the only value.
    assert list(cache.values())[0] == 1  # passes.

    # Replace the existing 'a' entry with a new value.
    cache['a'] = 200

    # __getitem__ works as expected.
    assert cache['a'] == 200  # passes.

    # Both dict and accessing via values() return the old entry: 1.
    assert dict(cache) == {'a': 200}  # fails.
    assert list(dict(cache).values())[0] == 200  # fails.
```

This test fails in the last asserts (of course to see the 2nd failure one needs to comment the first):

```pytest
λ pytest foo.py
======================== test session starts ========================
platform win32 -- Python 3.10.9, pytest-7.4.2, pluggy-1.3.0
rootdir: c:\Users\bruno\projects\boltons
configfile: pytest.ini
collected 1 item

foo.py F                                                       [100%]

============================= FAILURES ==============================
_____________________________ test_lru ______________________________

    def test_lru():
        from boltons.cacheutils import LRU

        cache = LRU()

        # Add an entry.
        cache['a'] = 1

        # Normal __getitem__ access.
        assert cache['a'] == 1  # passes.
        # Convert to dict.
        assert dict(cache) == {'a': 1}  # passes.
        # Another way to access the only value.
        assert list(cache.values())[0] == 1  # passes.

        # Replace the existing 'a' entry with a new value.
        cache['a'] = 200

        # __getitem__ works as expected.
        assert cache['a'] == 200  # passes.

        # Both dict and accessing via values() return the old entry: 1.
>       assert dict(cache) == {'a': 200}  # fails.
E       AssertionError: assert {'a': 1} == {'a': 200}
E         Differing items:
E         {'a': 1} != {'a': 200}
E         Use -v to get more diff

foo.py:23: AssertionError
====================== short test summary info ======================
[...]

PR #349, opened 2023-10-12
Test and fix for #348

mahmoud:
The test repro'd #348, plus I checked `.values()` for good measure. It wasn't really a supported use case before, but ideally it should be; cacheutil are dict subtypes (for better or worse). 

Before merging, I should probably add LRI tests for this, just to be safe. Also need to think a little harder about whether there are more invariant tests that need to be added for the linked list.

### DIFF
--- a/tests/test_cacheutils.py
+++ b/tests/test_cacheutils.py
@@ -167,6 +167,31 @@ def test_lru_basic():
     assert second_lru != lru
 
 
+def test_lru_dict_replacement():
+    # see issue #348
+    cache = LRU()
+
+    # Add an entry.
+    cache['a'] = 1
+
+    # Normal __getitem__ access.
+    assert cache['a'] == 1  # passes.
+    # Convert to dict.
+    assert dict(cache) == {'a': 1}  # passes.
+    # Another way to access the only value.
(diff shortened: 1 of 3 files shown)


tests/test_cacheutils.py
```diff
@@ -167,6 +167,31 @@ def test_lru_basic():
     assert second_lru != lru
 
 
+def test_lru_dict_replacement():
+    # see issue #348
+    cache = LRU()
```
nicoddemus, line 172: I think we can use the same test for both `LRU` and `LRI`.

```suggestion
@pytest.mark.parametrize("lru_class", [LRU, LRI])
def test_lru_dict_replacement(lru_class):
    # see issue #348
    cache = lru_class()
```

nicoddemus (approved):
Missing a CHANGELOG entry perhaps?

(Not sure how the changelog is updated in `boltons`, if during PR or at release time)

mahmoud:
Hey, thanks for your patience! @markrwilliams, one of the original authors also took a look and gave it a :+1: offline. Just a good old fashioned latent bug. As far as changelogs, I usually hammer those out at release. :) Thanks!

PR #350, opened 2023-10-29
Adding fill parameters to pairwise and windowed

mahmoud:
I found these handy for when I needed lookahead/lookbehind, but didn't want to add extra processing code before/after the processing loop. Here's a sketch of one way I've used it:

```python
for record, next_record in pairwise_iter(records, fill=None):
  if next_record is not None:
     record.link(next_record)
  record.process()
```

Without fill, I have to call `records[-1].process()` after the loop, assuming it's a list and not an iterable.

### DIFF
--- a/boltons/iterutils.py
+++ b/boltons/iterutils.py
@@ -64,7 +64,7 @@ except ImportError:
 
 try:
     from future_builtins import filter
-    from itertools import izip
+    from itertools import izip, izip_longest as zip_longest
     _IS_PY3 = False
 except ImportError:
     # Python 3 compat
@@ -72,6 +72,7 @@ except ImportError:
     basestring = (str, bytes)
     unicode = str
     izip, xrange = zip, range
+    from itertools import zip_longest
 
 
 def is_iterable(obj):
@@ -424,7 +425,7 @@ def chunk_ranges(input_size, chunk_size, input_offset=0, overlap_size=0, align=F
             return
 
 
-def pairwise(src):
+def pairwise(src, fill=_UNSET):
     """Convenience function for calling :func:`windowed` on *src*, with
     *size* set to 2.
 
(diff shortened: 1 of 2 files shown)


kurtbrose:
Makes sense.

Let me throw some ideas out there to see if there's any convenient improvements.

Is there a clearer name? `fill` makes me think there can be 0 to N items, like https://numpy.org/doc/stable/reference/generated/numpy.ndarray.fill.html or https://docs.python.org/3/library/stdtypes.html#str.zfill ;  In this case there will be exactly one value at the end.  Maybe `last` or `tail`?  (That also opens up `first` or `head` as a convenient name if you want to insert a special value for the first one).

This might be a highly simplified piece of code you are showing me, but some alternatives:

```python
for record, next_record in pairwise_iter(records + [None]):
   if next_record:
       record.link(next_record)
   record.process()
```

```python
for record, next_record in pairwise_iter(records):
    record.link(next_record)

for record in records:
    record.process()
```

mahmoud:
Haha, definitely highly simplified! But even so, first alternative eats up time/memory making a copy, and the second, **two** iterations?? We can do better! ;D 

In all seriousness, as I was modifying these functions for the first time in a decade, I got a little curious if the old `tees` approach used by `windowed` is particularly fast/efficient. Not curious enough to go down a benching rabbithole though :)

As far as naming, `fill` for sure makes sense for `windowed`, because it can occur up to `size - 1` times in the same window. For pairwise, I see your point, it can only really show up once. I guess `sentinel` or `terminator` or `end` could also work? I like `end` bc it's short :)

kurtbrose:
I like `end` too -- it's short and also makes it clear where this value will go.
Do you think there's anything special about the last item that doesn't apply to the first?
(That is, should we include a `start` at the same time as `end`?)

mahmoud:
@kurtbrose Ah, I get the symmetry with `start`, for `pairwise` usages of the form `(prev, cur)`, versus `(cur, nxt)`, like what I'm using above. I just haven't had a lot of those come up. And it feels maybe trivial to turn code from `(prev, cur)` to `(cur, nxt)`, right?

kurtbrose:
I don't have a use case for `start=` -- just to throw out a hypothetical, maybe you have parent pointers?

```python
for parent, child in pairwise_iter(nodes, start=None):
   child.parent = parent
```

you could do this the other way around:

```python
for child, parent in pairwise_iter(reversed(nodes), end=None):
    child.parent = parent
```

The hardest part is probably keeping `child, parent` versus `parent, child` consistent with `reversed().

PR #352, opened 2023-10-31
Quick refresh of the README

mahmoud:
Mostly checking to see if I can fix the MacPorts badge.


mahmoud:
MacPorts badge looks like it's working again. Seems like something that broke with the :apple: emoji + GitHub's caching usercontent system.

PR #353, opened 2023-11-01
Fix including docs and tests in sdist

mgorny:
Fix the `recursive-include` statements in `MANIFEST.in` to actually include full documentation and test tree.  The statements without a list of globs were ineffective, and only `.py` files were included. Effectively, docs were missing completely and tests missed required data files (e.g. `json_test_data.txt`).

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -4,5 +4,5 @@ include README.md
 include TODO.rst
 include pytest.ini
 include tox.ini
-recursive-include docs 
-recursive-include tests
+recursive-include docs *
+recursive-include tests *


mahmoud:
Ah, good ole MANIFEST :sweat_smile: 

Thanks for this!

mahmoud:
Just FYI, I did have to tweak this before 23.1.1, bc this was picking up pyc files and sphinx build side effects. TIL sphinx pickles the whole environment, making the boltons sdist >1MB :) 

Thanks again!

mgorny:
Thanks. If it turns out to pick too many files again, please let me know and I'll try to restrict the list a bit.

Issue #190, opened 2018-11-28
Adding type hints


wcooley:
I have started generating/creating typing hints for _boltons_ and would like to make them available for others to use. There are basically 3 ways that I can do that:

1. Add to the actual code (using Python legacy-compatible comments).
2. Add type stubs alongside the actual code.
3. Create a separate stub-only package (& repo).

Option 1 is the most invasive but also ensures greater consistency between the production code and hints and makes it more immediately available to someone reading the code.

Option 2 is less invasive but makes the hints less convenient for human consumption and adds a new part of the package to be maintained.

Option 3 is least invasive and requires no support from anyone other than myself but is quite likely to be neglected and poorly-maintained.

Any preference/comments?

wcooley:
Thanks for the thoughts @dmwyatt.

On one hand, I agree with you about the aesthetics of mixing typed & non-typed code; on the other hand, I have to admit that pretty much all of my code is like that -- I add typing in areas where I am working, but I have generally not added it wholesale to legacy code. Even new code I sometimes realize that I've left typing out, partly because PyCharm already does a lot of type inference, so it's absence is only noticeable where that fails and it gives me bogus warnings.

In the meantime, I have created a standalone stub distribution/package:

https://github.com/wcooley/python-boltons-stubs

You'll note that there is a lot of stuff missing. Some of this is due to lack of test coverage but in some cases MonkeyType failed to either collect/store the needed info or generate the stubs from what was collected.

I am attempting to use _mypy_ to find what is missing but I am having trouble getting it to work.

mahmoud:
Wow! Just how much of that repo was (pardon the pun) typed by hand? In any case, I appreciate the [typeshedding](https://github.com/python/typeshed/). I'll weigh the options and get back to you, but at the very least I'm certainly open to documenting this more publicly for those who might be interested in this usage pattern. Thanks again!

wcooley:
Thanks @mahmoud.
The typestubs were mostly generated with [MonkeyType](/instagram/MonkeyType) and a fixed-up [pytest-monkeytype](/mariusvniekerk/pytest-monkeytype).
There are a number of things without typestubs, because it \*ahem\* only generates stubs from code with test coverage.
I did have to manually generalize types, as it only generates typestubs based on the concrete types seen at runtime (which makes for interesting reading), but only for the types that I had immediate need for, so there is still much to be done.

kasium:
[...]

PR #356, opened 2023-11-29
Fix `boltons.strutils.__all__` missing names

### LINKED ISSUE #355: Names in `boltons.strutils.__all__` with no definitions
`boltons.strutils.__all__` contains `int_list_complement` and `int_list_to_int_tuples` which are not defined anywhere

Avasam:
Closes #355 

Missed in https://github.com/mahmoud/boltons/commit/17c64d01c2d9837a3b60f67de664de4a2cde7918, CC @mahmoud 

This could've been caught by a linter (see Flake8/Ruff) or a type-checker (see mypy/pyright, #190)


mahmoud:
Good catch, I'll keep that in mind, thanks!

PR #357, opened 2023-11-29
Use license_files instead of license_file in setup.cfg

bkmgit:
Use license_files instead

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -2,4 +2,4 @@
 universal = 1
 
 [metadata]
-license_file = LICENSE
+license_files = LICENSE


bkmgit:
See https://setuptools.pypa.io/en/latest/userguide/declarative_config.html#metadata

mahmoud:
Interesting, though that link doesn't specify that license_file is deprecated, just that it has a new alias. Is there another source you can point me to? Can you elaborate on why we'd introduce an implicit minimum version of setuptools (42.0.0)?

bkmgit:
This is a warning that came up when building for inclusion on Fedora linux. See the log:
https://download.copr.fedorainfracloud.org/results/@fedora-review/fedora-review-2252017-python-boltons/fedora-rawhide-x86_64/06705253-python-boltons/builder-live.log.gz

Is there a good way to support multiple versions of setuptools/python?

mahmoud:
Got it, thanks for the log. I've opened an issue on setuptools to update their docs: https://github.com/pypa/setuptools/issues/4147

As for version management, I think we're ok to upgrade. I looked up the setuptools version and any version released after 2019 should be fine. We could add setuptools>42 or something to `setup_requires`, but that'd mostly be as a form of documentation.

PR #362, opened 2024-02-03
Drop support for EOL Python <= 3.6

hugovk:
Fixes https://github.com/mahmoud/boltons/issues/339.

Also run https://github.com/asottile/pyupgrade to help upgrade to modern syntax.

PR #366, opened 2024-04-21
Remove more redundant internal compatibility aliases

Avasam:
Noticed in https://github.com/python/typeshed/pull/11804 . Follow-up to #362

idk if you'd like to drop tests for `total_ordering` since boltons is no longer implementing it.

### DIFF
--- a/boltons/deprutils.py
+++ b/boltons/deprutils.py
@@ -30,9 +30,8 @@
 
 
 import sys
+from types import ModuleType
 from warnings import warn
-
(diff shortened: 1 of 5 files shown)


mahmoud:
Hey @Avasam just circling back to this, this is great. A really good start to leaving the dust of 2 behind. Seems like the total_ordering tests at least work as a check that the symbol's there for backwards compat. Seems fast enough and pretty unlikely to change, right? Thanks again!

PR #374, opened 2024-09-19
Update variable output command in tests.yaml

rffontenelle:
Solves:

> The `set-output` command is deprecated and will be disabled soon. Please upgrade to using Environment Files. For more information see: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/


mahmoud:
Whoa, nice one! Thanks!

PR #376, opened 2024-12-03
add support for python 3.13

branchvincent:
This is mainly just declaring official support for 3.13

- Adds 3.13 to CI and bumps test deps due to [`ast` related deprecations](https://github.com/mahmoud/boltons/actions/runs/12073748189/job/33670473176#step:8:92)
- Replaces use of deprecated `utcnow` / `utcfromtimestamp`: https://github.com/python/cpython/issues/103857. Timezone-aware datetimes are now preferred,  but `isoparse` and `decimal_relative_time` currently  only support naive datetimes (no functional changes here)
- While here, migrate to [`pyproject.toml`](https://packaging.python.org/en/latest/guides/modernize-setup-py-project/), which supersedes `setup.py` and `MANIFEST.in`


mahmoud:
Hey Branch! This is great. Looks like there's a blocker on Windows CI. Something about a `getuser` failure, caused by `ModuleNotFoundError: No module named 'pwd'`. Might be due to something related to pytest's `tmp_path` fixture? 

If it's `tmp_path`, then it might be fixed by a pytest version bump. Feel free to upgrade dependencies; we can drop 3.7 if need be. If it's worse than that, or if it's easier, we can probably make our own, simpler temporary path fixture that won't hit the Windows `getuser` issue. Wdyt?

branchvincent:
> If it's tmp_path, then it might be fixed by a pytest version bump

Yup, you were spot on! Looks like we just need https://github.com/pytest-dev/pytest/commit/14d3707818e0beaab2a3ce4f4fa95ee5a147b75e. I updated to a more recent `pytest` version for 3.8+, so we can keep 3.7 support around 😄

PR #377, opened 2024-12-30
Fix incorrect usage of lstrip bug in fileutils tests

theartful:
I'm using Arch linux (typical starting statement of an arch user lol), and one of the packages I use depend on boltons (https://aur.archlinux.org/packages/python-boltons). I faced an issue with tests failing causing me to do some manual intervention.

On my machine, I have
```python
BOLTONS_PATH = "/home/theartful/.cache/yay/python-boltons/src/boltons/boltons"
```

The test checks that "fileutils.py" can be found using `iter_find_files`, and indeed it is part of `iter_find_files(BOLTONS_PATH, patterns=['*.py'])` as:
```python
p = "/home/theartful/.cache/yay/python-boltons/src/boltons/boltons/fileutils.py"
```

The bug here is that "lstrip" doesn't remove the prefix, but instead removes all characters from the left that is part of the string. For example:
```python
>>> BOLTONS_PATH = "/home/theartful/.cache/yay/python-boltons/src/boltons/boltons"
>>> p = "/home/theartful/.cache/yay/python-boltons/src/boltons/boltons/fileutils.py"
>>> p.lstrip(BOLTONS_PATH)
 ileutils.py
```

### DIFF
--- a/boltons/strutils.py
+++ b/boltons/strutils.py
@@ -57,7 +57,7 @@ __all__ = ['camel2under', 'under2camel', 'slugify', 'split_punct_ws',
            'iter_splitlines', 'indent', 'escape_shell_args',
            'args2cmd', 'args2sh', 'parse_int_list', 'format_int_list',
            'complement_int_list', 'int_ranges_from_int_list', 'MultiReplace',
-           'multi_replace', 'unwrap_text']
+           'multi_replace', 'unwrap_text', 'removeprefix']
 
 
 _punct_ws_str = string.punctuation + string.whitespace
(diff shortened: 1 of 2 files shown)


mahmoud:
Great catch, sincere thanks from a reformed Gentoo/Arch user. This is a classic bug, but we have to do a bit more to make the tests pass on older Pythons and Windows.

Here's a removeprefix that you can just drop in the file and use, or add to strutils if you're so inclined:

```
def removeprefix(text: str, prefix: str) -> str:
    if text.startswith(prefix):
        return text[len(prefix):]
    return text
```

As for Windows, the file path separator is throwing off tests, so we still need to strip that out with `.lstrip(os.path.sep)`

Looking forward to merging this, thanks again!

theartful:
I added `removeprefix` to `strutils.py`. I forgot all about windows and their usage of backslashes instead of forward slashes.

mahmoud:
Looks great! Thanks so much again!

PR #381, opened 2025-02-07
Fix Table.to_text() sizing

osoleve:
Table.to_text now measures column contents instead of row length when sizing the individual columns.

Before:
image

After:
image


mahmoud:
Great catch, thanks for this!

PR #388, opened 2025-09-28
feat(strutils): add human_readable_list function to format lists of strings

gtkacz:
Also formatted the imports according to PEP 8


mahmoud:
Hey there! Love the idea, love that oxford is True by default, but looks like there's a syntax error?

gtkacz:
My bad, I was using f-strings as first class objects, but that was only introduced in PEP 701 for Python 3.12. Pushing the fix now

PR #389, opened 2025-10-04
Fix barrellist sort with multiple lists

Matthieu-Beauchamp:
The following code in `BarrelList.sort` causes a crash

```py
del self.lists[:]
self.lists[0] = tmp_sorted
```


mahmoud:
Nice fix, with a test, no less. LGTM, thanks!

PR #400, opened 2026-04-18
Fix syntax error in MultiReplace example

BartoszCki:


mahmoud:
Appreciate this, good eye!

PR #404, opened 2026-06-17
Fix OrderedMultiDict equality to compare values against plain mappings

DucMinhNe:
OrderedMultiDict.__eq__ has a branch for comparing against a non-OMD mapping (anything with a keys() method, like a dict). That branch computes other[selfk] == self[selfk] but discards the result, so it only checks that every key of self exists in other and never compares the values. An OMD therefore compares equal to any same-keyed mapping no matter what the values are:

```python
>>> from boltons.dictutils import OMD
>>> omd = OMD([("a", 1), ("b", 2)])
>>> omd == {"a": 999, "b": 2}
True   # should be False
```

The existing test_eq does not catch this because its dict case is written as `d == omd`, which dispatches to `dict.__eq__` (OMD subclasses dict), not `OMD.__eq__`. The buggy branch is only reached with the OMD on the left.

Fix: compare the values and return False on mismatch. The length check earlier in __eq__ already handles the differing-key-count case.

The same __eq__ is duplicated in urlutils.OrderedMultiDict (the base class of QueryParamDict), which had the identical bug, so it is fixed there too.

Added regression tests in test_dictutils.py and test_urlutils.py that put the OMD/QueryParamDict on the left so the branch is actually exercised. Both fail before the change and pass after. Full suite: 433 passed.

### DIFF
--- a/boltons/dictutils.py
+++ b/boltons/dictutils.py
@@ -358,7 +358,8 @@ class OrderedMultiDict(dict):
         elif hasattr(other, 'keys'):
             for selfk in self:
                 try:
-                    other[selfk] == self[selfk]
+                    if other[selfk] != self[selfk]:
+                        return False
                 except KeyError:
                     return False
             return True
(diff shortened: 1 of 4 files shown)


mahmoud:
Wow, that is a _big_ miss and a great catch. Good call, thank you!

PR #406, opened 2026-06-19
Fix split_iter maxsplit=0 handling

CodingFeng101:
## Summary

Fix `split_iter(..., maxsplit=0)` so it yields the unsplit input values instead of wrapping the source iterable object itself.

Previously, the `maxsplit == 0` branch yielded `[src]`. For list inputs this produced an extra nested list, and for iterator inputs it returned the iterator object rather than its values.

## Validation

- Added a regression test covering both list and iterator inputs
- Confirmed the new test failed before the fix with `[[[1, None, 2]]] != [[1, None, 2]]`
- Ran `python -m pytest tests/test_iterutils.py::TestSplit::test_maxsplit_zero_returns_unsplit_values tests/test_iterutils.py -q` and confirmed 44 passed
- Ran `python -m pytest -q` and confirmed 436 passed

### DIFF
--- a/boltons/iterutils.py
+++ b/boltons/iterutils.py
@@ -158,7 +158,7 @@ def split_iter(src, sep=None, maxsplit=None):
     if maxsplit is not None:
         maxsplit = int(maxsplit)
         if maxsplit == 0:
-            yield [src]
+            yield list(src)
             return
 
     if callable(sep):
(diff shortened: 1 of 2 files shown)


mahmoud:
Oh wow, good call on an edge case. Thanks for the fix + test!

PR #407, opened 2026-06-19
Match list insert before start in BarrelList

CodingFeng101:
## Summary

Make `BarrelList.insert()` match built-in `list.insert()` when the insertion index is before the start of a multi-bucket `BarrelList`.

`BarrelList` is documented as a `list` subtype with an identical API. Built-in lists clamp very negative insert indexes to the beginning, but `BarrelList` raised `IndexError` after it had been balanced into multiple internal lists.

## Validation

- Added a regression test for inserting with an index far below `-len(bl)` after balancing
- Confirmed the new test failed before the fix with `IndexError`
- Ran `python -m pytest tests/test_listutils.py::test_barrel_list_insert_before_start_matches_list tests/test_listutils.py -q` and confirmed 4 passed
- Ran `python -m pytest -q` and confirmed 436 passed

### DIFF
--- a/boltons/listutils.py
+++ b/boltons/listutils.py
@@ -139,7 +139,7 @@ class BarrelList(list):
         else:
             list_idx, rel_idx = self._translate_index(index)
             if list_idx is None:
-                raise IndexError()
+                list_idx, rel_idx = 0, 0
             self.lists[list_idx].insert(rel_idx, item)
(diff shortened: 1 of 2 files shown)


mahmoud:
Another tricky corner case, nice one. Thanks!

PR #411, opened 2026-06-22
tbutils: preserve ParsedException anchor lines

### LINKED ISSUE #333: `ParsedException.from_string(text).to_string() == text` property violated due to anchors
This traceback
```
  File "main.py", line 3, in <module>
    print(add(1, "two"))
          ^^^^^^^^^^^^^
  File "add.py", line 2, in add
    return a + b
           ~~^~~
TypeError: unsupported operand type(s) for +: 'int' and 'str'
```
as an example includes anchors (the lines with "~" and "^"). In a [separate PR](https://github.com/mahmoud/boltons/pull/332), ability to parse tracebacks with these was added, but information about column number is not stored in the ParsedException object in order to reconstruct this.

To maintain the property that `ParsedException.from_string(text).to_string() == text`, it will be necessary to add an attribute to the `ParsedException` class, and the logic in both `from_string` for parsing and `to_string` for reconstruction.

gaoflow:
## Summary
- store traceback source anchor lines on parsed frames as `source_line_anchor`
- emit stored anchors from `ParsedException.to_string()`
- update the anchor regression test to assert exact `from_string(...).to_string()` round-tripping

Closes #333.

## Validation
- `python -m pytest tests/test_tbutils_parsed_exc.py::test_parsed_exc_with_anchor -q`
- `python -m pytest tests/test_tbutils_parsed_exc.py tests/test_tbutils.py -q`
- `python -m pytest -q` (437 passed)
- `python -m compileall boltons/tbutils.py tests/test_tbutils_parsed_exc.py`
- `git diff --check`

This pull request was prepared with the assistance of AI, under my direction and review.

### DIFF
--- a/boltons/tbutils.py
+++ b/boltons/tbutils.py
@@ -716,9 +716,9 @@ class ParsedException:
 
         .. note::
 
-           Note that this method does not output "anchors" (e.g.,
-           ``~~~~~^^``), as were added in Python 3.13. See the built-in
-           ``traceback`` module if these are necessary.
+           Note that this method stores traceback anchor lines (e.g.,
+           ``~~~~~^^``), but does not store SyntaxError details as
+           structured fields.
         """
         lines = ['Traceback (most recent call last):']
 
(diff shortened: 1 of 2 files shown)


mahmoud:
Appreciate the report, the #333 round-trip gap is real. When I dug in, the fix hardcoded the 4-space dedent (`next_line[4:]`), which loses `from_string`'s tolerance for indented tracebacks, so I knocked it out in #425. Thanks!

Issue #261, opened 2020-07-07
Python `funcutils.wraps` breaks when keyword only arguments involved


thorwhalen:
Code worth 1000 words:

```python
>>> from boltons.funcutils import wraps
>>>
>>> def g(a: float, b=10):
...     return a * b
>>>
>>> def f(a: int,  *, b=1):
...     return a * b
>>>
>>> # all is well here...
>>> assert f(3) == 3
>>> assert g(3) == 30
>>> assert wraps(f)(g)(3) == 3  # yay, g got the f default (not so with functools.wraps!)
>>>
>>> # but this doesn't work
>>> wraps(g)(f)(3)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "<boltons.funcutils.FunctionBuilder-1>", line 2, in g
TypeError: f() takes 1 positional argument but 2 were given
```

mahmoud:
Ah, I was confused when this bug came in, bc I've got scars from dealing with the expansion of function signature semantics. 

You'll notice that when kw-only args are involved the first time (`wraps(f)(g)`) it works fine. The other way around, however, `wraps(g)(f)` fails, because by the very nature of `funcutils.wraps`, you're trying to give `f` the same signature as `g`. And `f`'s signature is not like `g`'s. `g` supports positional `b` and `f()` does not.

If you want to make a wrapper which converts a keyword-only argument to one that can be positional or keyword only, that'll require a different approach for now.

A potential fix would be to pass all function arguments as keywords. But doubt that's the right direction, because, while I have yet to add positional argument only support, that'll definitely throw a wrench into things.

thorwhalen:
We got there eventually.

PR #412, opened 2026-06-24
funcutils: forward defaulted args as keyword args in wraps invocation

### LINKED ISSUE #343: wraps loses keywords
Thanks for the great project!

The decorator `boltons.funcutils.wraps` seems to lose keywords passed to the wrapped function.  This does not happen with `functools.wraps`.
Here is a silly example:

```python

def flip(f):
    @wraps(f)
    def wrapper(*args, **kwargs):
        return f(*reversed(args), **kwargs)
    return wrapper

def pow(x, y, msg=""):
    """Print msg and raise x to the power of y."""
    print(f"{x=}, {y=}")
    if msg:
        print(f"{msg=}")
    result = x ** y
    print(f"{result=}")
    return result

flipped_pow = flip(pow)
flipped_pow(3, 2, msg="abc")
```
---
Expected:
> x=2, y=3
> msg='abc'
> result=8
---
Actual:
> x='abc', y=2
> msg=3

<details>
<summary>TypeError</summary>

```python
TypeError                                 Traceback (most recent call last)
Cell In[22], line 18
     15     return x ** y
     17 flipped_pow = flip(pow)
---> 18 flipped_pow(3, 2, msg="abc")

File :2, in pow(x, y, msg)

Cell In[22], line 7, in flip..wrapper(*args, **kwargs)
      5 @wraps(f)
      6 def wrapper(*args, **kwargs):
----> 7     return f(*reversed(args), **kwargs)

Cell In[22], line 15, in pow(x, y, msg)
     13 if msg:
     14     print(f"{msg=}")
---> 15 return x ** y

TypeError: unsupported operand type(s) for ** or pow(): 'str' and 'int'
```

</details>

---
I also noticed that if I make msg keyword only, as below, it works:
```python
def pow(x, y, *, msg=""):
```

gaoflow:
Fixes #343.

## The bug

`FunctionBuilder.get_invocation_str()` passes every regular argument
positionally in the generated body:

```python
# generated for:  def pow(x, y, msg=''):
return _call(x, y, msg)
```

When a caller supplies a defaulted parameter by keyword (`pow(3, 2, msg='abc')`),
the generated function correctly binds `msg='abc'`, but then forwards it
positionally as `_call(3, 2, 'abc')`.  Any wrapper that inspects
`*args` / `**kwargs` separately—the standard pattern—loses the keyword
information:

```python
def flip(f):
    @wraps(f)
    def wrapper(*args, **kwargs):
        return f(*reversed(args), **kwargs)
    return wrapper

def power(x, y, msg=''):
    return (x, y, msg)

flip(power)(3, 2, msg='abc')
# before: ('abc', 2, 3)  – msg landed in args, was reversed
# after:  (2, 3, 'abc')  – correct
```

## The fix

Args that carry a default are now forwarded with the `name=name` form:

```python
return _call(x, y, msg=msg)
```

This means the inner wrapper receives them in `**kwargs`, preserving
exactly how the caller passed them.

As a welcome side-effect, this also lifts the limitation documented in
`test_wraps_inner_kwarg_only` (issue #261): wrapping a keyword-only-`b`
function with a signature that exposes `b` as positional-or-keyword now
works, because the generated invocation passes `b=b`.

## Test changes

- Added `test_wraps_preserves_kwargs_for_defaulted_args` in
  `test_funcutils.py` directly covering the issue #343 scenario.
- `test_wraps_inner_kwarg_only`: updated docstring and assertion to
  reflect that `wraps(g)(f)(3)` now succeeds (was asserting `TypeError`
  which documented the old limitation).
- `test_wraps_expected` (`expect_pair` / `expect_dict`): these wrappers
  extracted defaulted injected args from `args[-1]`, relying on the old
  positional-passthrough. Updated to use `kwargs.pop(...)`, which is the
  correct pattern under the new semantics.
- Parametrized `test_get_invocation_sig_str`: updated expected invocation
  string for the single-defaulted-arg case (`"a"` → `"a=a"`).

All 438 tests pass.

### DIFF
--- a/boltons/funcutils.py
+++ b/boltons/funcutils.py
@@ -790,23 +790,28 @@ class FunctionBuilder:
     """, re.VERBOSE)
 
     def get_invocation_str(self):
-        kwonly_pairs = None
-        formatters = {}
-        if self.kwonlyargs:
-            kwonly_pairs = {arg: arg
-                                for arg in self.kwonlyargs}
-            formatters['formatvalue'] = lambda value: '=' + value
+        # Build the invocation string that forwards all arguments to _call.
+        # Regular args that have defaults must be forwarded as keyword
+        # arguments (``name=name``) so that callers who supply them by
+        # keyword have those values arrive in *kwargs* inside the wrapper,
+        # not silently flattened into a positional tuple.  Without this,
+        # ``wrapper(*args, **kwargs)`` loses keyword information for
+        # defaulted parameters (issue #343).
+        defaults = self.defaults or ()
+        n_positional = len(self.args) - len(defaults)
+        positional_args = self.args[:n_positional]
+        defaulted_args = self.args[n_positional:]
 
-        sig = inspect_formatargspec(self.args,
-                                    self.varargs,
-                                    self.varkw,
-                                    [],
-                                    kwonly_pairs,
-                                    kwonly_pairs,
-                                    {},
-                                    **formatters)
-        sig = self._KWONLY_MARKER.sub('', sig)
-        return sig[1:-1]
+        parts = list(positional_args)
+        if self.varargs:
+            parts.append('*' + self.varargs)
+        for arg in defaulted_args:
+            parts.append(arg + '=' + arg)
+        for arg in (self.kwonlyargs or []):
+            parts.append(arg + '=' + arg)
+        if self.varkw:
+            parts.append('**' + self.varkw)
+        return ', '.join(parts)
 
     @classmethod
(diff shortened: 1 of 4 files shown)


mahmoud:
Appreciate the dig on #343, keyword-forwarding is the right direction. As written though it breaks positional-only params (`def f(x, y=2, /)` raises TypeError) and defaulted args before `*args` (`f(x, y=1, *va)` gets 'multiple values for argument'), so I took my own pass in #425. Thanks!

PR #413, opened 2026-06-29
Handle bytes from socket.gethostname() in GUIDerator.reseed

### LINKED ISSUE #295: Default value for socket.gethostname() should be str, not bytes
https://github.com/mahmoud/boltons/blob/270e974975984f662f998c8f6eb0ebebd964de82/boltons/iterutils.py#L1322-L1332

If the default value is used, `'-'.join([...])` fails because the second item is bytes, not str.

uttam12331:
## Summary
Closes #295.

On some platforms/configurations `socket.gethostname()` returns `bytes` rather than `str`. `GUIDerator.reseed()` joins it into the salt with `str.join()`:

```python
self.salt = '-'.join([str(self.pid),
                      socket.gethostname() or '<nohostname>',
                      ...])
```

which raises `TypeError: sequence item 1: expected str instance, bytes found`. Because a module-level `GUIDerator()` is created at import time, this can break importing `boltons.iterutils` entirely on affected systems.

## Fix
Decode the hostname to `str` when it comes back as `bytes`:

```python
hostname = socket.gethostname()
if isinstance(hostname, bytes):
    hostname = hostname.decode('utf8', 'replace')
```

## Tests
Added `test_guiderator_bytes_hostname` (monkeypatches `socket.gethostname` to return `b'myhost'`) asserting a valid hex GUID is produced without raising.

Verified locally: `pytest tests/test_iterutils.py` → **45 passed**.

### DIFF
--- a/boltons/iterutils.py
+++ b/boltons/iterutils.py
@@ -1475,8 +1475,13 @@ class GUIDerator:
     def reseed(self):
         import socket
         self.pid = os.getpid()
+        hostname = socket.gethostname()
+        # gethostname() may return bytes on some platforms/configs; decode so
+        # the str.join() below does not raise a TypeError on a mixed sequence.
+        if isinstance(hostname, bytes):
+            hostname = hostname.decode('utf8', 'replace')
         self.salt = '-'.join([str(self.pid),
-                              socket.gethostname() or '<nohostname>',
+                              hostname or '<nohostname>',
                               str(time.time()),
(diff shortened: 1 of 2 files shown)


mahmoud:
Thanks, but I don't believe this one's real: CPython's `socket.gethostname()` always returns str. #295 was about the old bytes *default* literal, which was fixed in 65e0354. Closing.

PR #414, opened 2026-06-29
test(pathutils): add unit tests for augpath, shrinkuser, expandpath

LD-RW:
## What

Adds `tests/test_pathutils.py` — Arrange/Act/Assert unit tests for the three
functions in `boltons/pathutils.py` (`augpath`, `shrinkuser`, `expandpath`).

## Why

`pathutils` is the only major `*utils` module without a dedicated test file.
The module's doctests cover the headline examples, but two things motivated a
real test file:

- Several parameter combinations and edge cases aren't exercised anywhere:
  `augpath` directory preservation/replacement, missing vs empty vs replaced
  extensions, `multidot` true/false; `shrinkuser`'s no-separator guard and
  custom home symbol on the exact-home branch; `expandpath`'s undefined-variable
  passthrough.
- `--doctest-modules` is only applied by the tox command, not by
  `[tool.pytest.ini_options]`, so a plain `pytest` run collects no doctests.
  These unit tests give the module coverage independent of how the suite is run.

## Notes

- Tests only — no changes to `pathutils.py` or any other source.
- All home-directory assertions derive from `expanduser('~')` and paths are
  built with `os.path.join`, so the suite is machine- and platform-independent
  (relevant given the Windows/macOS CI matrix and the module's `%USERPROFILE%`
  docs).
- House style: plain functions with `assert`, matching the existing
  `tests/test_*utils.py` files. Env-var test uses the `monkeypatch` fixture.

Ran locally: 14 unit tests pass, and `pytest --doctest-modules
boltons/pathutils.py tests/test_pathutils.py` passes (doctests + unit tests
coexist).

### DIFF
--- a/tests/test_pathutils.py
+++ b/tests/test_pathutils.py
@@ -1,0 +1,272 @@
+"""Tests for boltons.pathutils.
+
+The functions in ``pathutils`` ship with doctests covering their headline
+examples, but the module had no dedicated unit-test file like the other
+``*utils`` modules do. These tests focus on the parameter combinations and
+edge cases the doctests do not exercise, following an Arrange/Act/Assert
+layout.
+
+All home-directory assertions are derived from ``expanduser('~')`` so the
+suite is independent of the machine it runs on.
+"""
+
+from os.path import expanduser, normpath, join
+
+from boltons.pathutils import augpath, shrinkuser, expandpath
+
+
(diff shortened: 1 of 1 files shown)


mahmoud:
Consolidated a leaner version of these into #425. Appreciate the nudge on pathutils coverage, thanks!

PR #416, opened 2026-06-30
fix(fileutils): accept os.PathLike in AtomicSaver

### LINKED ISSUE #383: Supporting more PathLike in path-related methods
Whilst working on https://github.com/mahmoud/boltons/pull/382 I noticed a few places (`AtomicSaver`'s `dest_path` parameter to name one) where the code could be updated to support `os.PathLike` instead of just a string.

uttam12331:
`AtomicSaver.__init__` assigned `dest_path` directly to `self.dest_path` and later concatenated it as a string (`dest_path + '.part'`), which raises `TypeError` when a `pathlib.Path` is passed.

`iter_find_files` called `.split()` on the `directory` argument, which also fails for `pathlib.Path`.

**Changes:**
- `AtomicSaver.__init__`: convert `dest_path` via `os.fspath()` at the top of the method, and use `self.dest_path` (the already-converted str) when building the `.part` path
- `iter_find_files`: convert `directory` via `os.fspath()` before calling `.split()`
- Add `test_atomicsaver_pathlike` and `test_iter_find_files_pathlike` to the test suite

Closes #383

### DIFF
--- a/boltons/fileutils.py
+++ b/boltons/fileutils.py
@@ -386,7 +386,7 @@ class AtomicSaver:
 
     # TODO: option to abort if target file modify date has changed since start?
     def __init__(self, dest_path, **kwargs):
-        self.dest_path = dest_path
+        self.dest_path = os.fspath(dest_path)
         self.overwrite = kwargs.pop('overwrite', True)
         self.file_perms = kwargs.pop('file_perms', None)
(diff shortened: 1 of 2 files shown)


mahmoud:
Good call, pathlib support here was overdue. Thanks for the fix + tests!

PR #417, opened 2026-07-02
fix: handle unencoded ? and / in URL credentials in parse_url()

### LINKED ISSUE #309: boltons.urlutils.URL cannot handle some characters in credentials
When trying to parse a URL with special characters in the credentials part it is not working as expected, I found an issue with `?` and `/`.

If it is in the password there is an exception
```
In [52]: URL("http://username:password?@www.proxy.com:443")
ValueError                                Traceback (most recent call last)
~/tests/venv/lib/python3.7/site-packages/boltons/urlutils.py in parse_url(url_text)
    938             try:
--> 939                 port = int(port_str)
    940             except ValueError:

ValueError: invalid literal for int() with base 10: 'password'

During handling of the above exception, another exception occurred:

URLParseError                             Traceback (most recent call last)
<ipython-input-52-4078cba2692c> in <module>
----> 1 URL("http://username:password?@www.proxy.com:443")

~/tests/venv/lib/python3.7/site-packages/boltons/urlutils.py in __init__(self, url)
    496                                         ' passing the result. (got: %s)'
    497                                         % (DEFAULT_ENCODING, ude))
--> 498             ud = parse_url(url)
    499
    500         _e = u''

~/tests/venv/lib/python3.7/site-packages/boltons/urlutils.py in parse_url(url_text)
    941                 if port_str:  # empty ports ok according to RFC 3986 6.2.3
    942                     raise URLParseError('expected integer for port, not %r'
--> 943                                         % port_str)
    944                 port = None
    945

URLParseError: expected integer for port, not 'password'
```

If it is in the username it is not raising an exception but the pairing is incorrect
```
In [56]: print(URL("http://username?:password@www.proxy.com:443").port)
None
```

This seems to be an issue with `urlparse` so I'm not sure it is `boltons` to blame.

Trying to use a regex pattern is working for me
```
pattern = re.compile(
r"""
    (?P<schema>[\w\+]+)://
    (?:
        (?P<username>[^:/]*)
        (?::(?P<password>.*))?
    @)?
    (?:
        (?:
            (?P<host>[^/:]+)
        )?
        (?::(?P<port>[^/]*))?
    )?
    """,
re.X,
)
```
(This still not working if the username contain `/`)

koteshyelamati:
Fixes #309

When a URL password contains an unencoded `?` or `/`, the URL regex stops the authority capture early (at the special character), absorbing the `@` credential separator into the query or path instead.

**Root cause:** The regex `[^/?#]*` for the authority group halts at `?`, `/`, or `#`. So `http://user:p?ss@host:443` is parsed as authority=`user:p`, query=`ss@host:443` — and then `parse_url()` finds no `@` in the authority and fails when it tries to parse `p` as a port.

**Fix:** After the normal `rpartition('@')` finds no `@` in the authority, detect two recovery cases:
- `query.startswith('@')` → the password had an unencoded `?`; reconstruct `userinfo = authority + '?'` and `hostinfo = query[1:]`
- `path.startswith('/@')` → the password had an unencoded `/`; reconstruct `userinfo = authority + '/'` and `hostinfo = path[2:]`

**Note:** Per RFC 3986, special characters in credentials should be percent-encoded (`?` → `%3F`, `/` → `%2F`). This fix makes boltons tolerant of the common real-world case where they are not.

### DIFF
--- a/boltons/urlutils.py
+++ b/boltons/urlutils.py
@@ -917,6 +917,29 @@ def parse_url(url_text):
         if sep:
             # TODO: empty userinfo error?
             user, _, pw = userinfo.partition(':')
+        else:
+            # No '@' found in authority. This can happen when the password
+            # contains an unencoded '?' or '/', causing the URL regex to stop
+            # the authority capture early and absorb the '@' into the query
+            # or path instead. Recover by detecting '@' at the start of
+            # the query (unencoded '?') or after '/' at the start of the
+            # path (unencoded '/'), then reconstruct userinfo and hostinfo.
+            query_text = gs.get('query') or ''
+            path_text = gs.get('path') or ''
+            if query_text.startswith('@'):
+                # Unencoded '?' in password: the regex treated '?' as the
+                # query delimiter and '@' landed at the start of the query.
+                userinfo = au_text + '?'
+                hostinfo = query_text[1:]
+                gs['query'] = None
+                user, _, pw = userinfo.partition(':')
+            elif path_text.startswith('/@'):
+                # Unencoded '/' in password: the regex treated '/' as the
+                # path start and '@' landed immediately after in the path.
+                userinfo = au_text + '/'
+                hostinfo = path_text[2:]
+                gs['path'] = ''
+                user, _, pw = userinfo.partition(':')
 
     host, port = None, None
     if hostinfo:


eeshsaxena (commented):
Thanks for digging into #309 - the root-cause analysis of the authority regex halting at `?` / `/` is accurate and well explained. I tried the patch out locally though, and I think the recovery heuristic has two problems that would block it as written. Sharing my reproduction so you can check my work.

I applied the diff to a copy of `urlutils.py` and compared `parse_url` before/after.

### 1. It silently mis-parses common URLs that have no credentials at all

The `path.startswith('/@')` branch fires on any URL whose path begins with `/@`, which is exactly the shape of a Mastodon profile link and an npm scoped package:

```
https://mastodon.social/@Gargron
  before:  host='mastodon.social'  username=None  path='/@Gargron'
  after:   host='Gargron'          username='mastodon.social/'  path=''

https://registry.npmjs.org/@babel/core
  before:  host='registry.npmjs.org'  username=None  path='/@babel/core'
  after:   host='babel/core'          username='registry.npmjs.org/'  path=''
```

The `query.startswith('@')` branch has the same issue:

```
http://example.com/?@handle
  before:  host='example.com'  query='@handle'
  after:   host='handle'       username='example.com?'  query=None
```

In each case the host is silently replaced by a fragment of the path, and the real host is moved into `username`. Because `parse_url` backs `URL`, anything doing `URL(u).host` for an allowlist or SSRF check would now read an attacker-influenced value. That makes this a bit riskier than a parsing nit.

### 2. It does not fix the case from the issue

The recovery only triggers when the offending character is the **last** character of the password, because only then does `@` land at position 0 of the query / right after `/`. The example in your PR description still raises:

```python
>>> parse_url('http://user:p?ss@host:443')
URLParseError: expected integer for port, not 'p'      # unchanged, with the patch applied
>>> parse_url('http://user:p/ss@host:443')
URLParseError: expected integer for port, not 'p'      # unchanged, with the patch applied
```

Only the narrower `http://user:pass?@host:443` form (the literal one in #309) starts working. So the patch handles the issue's exact string but not the general case it describes.

### Suggestion

I think the fix has to happen at the regex/authority-splitting layer rather than as a post-hoc recovery, since by the time the groups are captured the information about where the authority really ended is gone. One approach: before matching, find the *last* `@` that precedes the first `/`, `?` or `#` **that follows it**, and split the authority there explicitly - i.e. locate the credential separator first, then let the regex parse each side. That handles `p?ss`, `p/ss`, and multiple special characters, and it never fires when there is no `@` in the pre-path region, so `/@Gargron` stays untouched.

Whatever direction you take, it would be worth adding tests for the no-credential cases above (`/@user` paths, `?@` queries) alongside the credential ones - the current diff has no test coverage, which is why the regressions do not show up in CI.

Worth noting that per RFC 3986 `?` and `/` are genuinely not allowed unescaped in userinfo, so it is also a legitimate outcome for `parse_url` to keep rejecting them - the maintainers may prefer a clearer `URLParseError` over tolerance here. I will defer to @mahmoud on whether tolerance is wanted at all.

mahmoud:
Thanks for digging into #309, but I don't think this recovery can be made safe: `@` is legal in query strings and paths, so e.g. `http://example.com/@user/profile` would get parsed as credentials (`userinfo='example.com/'`, `host='user/profile'`). Unencoded `?`/`/` in userinfo isn't recoverable without misparsing valid URLs. Percent-encoding (%3F, %2F) is the way. Closing.

PR #420, opened 2026-07-09
fix(urlutils): QueryParamDict equality ignores key insertion order

### LINKED ISSUE #280: URL comparison is order-based for query parameters
Hi,

I am using `boltons` to do URL comparisons, being agnostic of the URL representation, and it works great so far except for query parameters.

Typically, with percent encoding, this is working as expected:

```python
>>> a = boltons.urlutils.URL('http://example.com/foo, bar')
>>> b = boltons.urlutils.URL('http://example.com/foo%2C%20bar')
>>> a == b
True
```

However, when taking query parameters into account, the comparison seems to be ordered (while I would expect the order of query parameters to have no impact). See:

```python
>>> a = boltons.urlutils.URL('http://example.com/foo, bar?foo=bar&bar=foo')
>>> b = boltons.urlutils.URL('http://example.com/foo%2C%20bar?bar=foo&foo=bar')
>>> a == b
False
```

I solved the issue with a bit of monkey patching, but is there a reason for having an ordered comparison here?

My solution:

```python
def query_params_dict_eq_monkey_patch(self, other):
    """
    >>> a = boltons.urlutils.URL('http://example.com/foo, bar?foo=bar&bar=foo')
    >>> b = boltons.urlutils.URL('http://example.com/foo%2C%20bar?bar=foo&foo=bar')
    >>> a == b
    True
    """
    return dict(self) == dict(other)
boltons.urlutils.QueryParamDict.__eq__ = query_params_dict_eq_monkey_patch
```

Thanks!

amitmishra11:
## Summary

`URL.__eq__` compares `query_params` using `OrderedMultiDict.__eq__`, which is insertion-order-sensitive. As a result, two URLs with identical query parameters in different order compare as unequal, even though RFC 3986 does not specify a canonical ordering for query parameters.

```python
a = URL('http://example.com/foo?foo=bar&bar=foo')
b = URL('http://example.com/foo?bar=foo&foo=bar')
a == b  # False  <-- was broken, now True
```

## Changes

- Add `QueryParamDict.__eq__` that compares by key set and per-key value lists, making key insertion order irrelevant while preserving the significance of value ordering within a repeated key (e.g. `k=1&k=2` vs `k=2&k=1` are still considered different).
- Add `QueryParamDict.__ne__` to match.
- Add six tests covering the new behaviour.

## Testing

```
pytest tests/test_urlutils.py  # 131 passed
```

Fixes #280

### DIFF
--- a/boltons/urlutils.py
+++ b/boltons/urlutils.py
@@ -1586,6 +1586,32 @@ class QueryParamDict(OrderedMultiDict):
         pairs = parse_qsl(query_string, keep_blank_values=True)
         return cls(pairs)
 
+    def __eq__(self, other):
+        """Compare two :class:`QueryParamDict` instances in a key-order-independent
+        way.
+
+        Two query param dicts are equal when they contain the same keys and,
+        for each key, the same list of values in the same order.  The
+        insertion order of *different* keys does not affect equality, which
+        matches the behaviour described in RFC 3986 §6.2.3.
+
+        >>> QueryParamDict.from_text('a=1&b=2') == QueryParamDict.from_text('b=2&a=1')
+        True
+        >>> QueryParamDict.from_text('a=1&a=2') == QueryParamDict.from_text('a=2&a=1')
+        False
+        """
+        if not isinstance(other, QueryParamDict):
+            return super().__eq__(other)
+        if set(self.keys()) != set(other.keys()):
+            return False
+        for key in self.keys():
+            if self.getlist(key) != other.getlist(key):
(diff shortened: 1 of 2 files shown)


mahmoud:
As discussed in #280, query param order is significant per the standard, and this `__eq__` would also be asymmetric against plain OMDs (order-insensitive QPD-to-QPD, order-sensitive QPD-to-OMD). Subclassing per #280 remains the recommended approach. Thanks anyway!

PR #423, opened 2026-07-17
Fix IndexedSet slicing returning wrong items after a removal

chuenchen309:
## The bug

`IndexedSet` advertises `list`-like indexing and slicing, but slicing returns the wrong items once anything has been removed:

```python
>>> from boltons.setutils import IndexedSet
>>> x = IndexedSet(range(10)); x.pop(2)   # x is now [0, 1, 3, 4, 5, 6, 7, 8, 9]
>>> list(x[1:4])
[1, 3, 4, 5]        # wrong — expected [1, 3, 4]
>>> list(x[-3:])
[8, 9]              # wrong — expected [7, 8, 9]
```

Silent wrong results, no exception. A freshly-built set with the same contents (no dead slots) slices correctly, so the bug only shows up after a removal.

## Cause

`iter_slice` runs `islice` over `self`, whose `__iter__` yields the **apparent** stream (dead slots skipped). But it first pushes `start`/`stop` through `_get_real_index()`, which maps an apparent index into `item_list` space by adding the dead slots that precede it. So `islice` gets `item_list`-space bounds while iterating the already-compacted apparent stream, over-counting by the removed slots before each bound.

## Fix

Keep the bounds in apparent space and normalize negatives the way slicing does (`len(self) + i`, floored at 0) instead of running them through `_get_real_index()`. For a set with no removals the behaviour is unchanged.

Verified with a fuzz comparison against `list` slicing — 5,000+ (start, stop, step) combinations across several removal patterns, 0 mismatches — and the existing `tests/test_setutils.py` stays green. Added a regression test covering the reported cases plus the exhaustive agreement.

Scope: negative-*step* slicing (`x[8:2:-1]`) is separately unimplemented (returns `[]` even with no removals); I left that untouched to keep this focused on the removal regression — happy to follow up if you'd like.

---
*Disclosure: prepared with AI assistance (Claude Code, Claude Opus 4.8). I reproduced the bug, wrote the failing regression test first, fuzz-verified the fix against `list` slicing, and reviewed the diff before opening.*

### DIFF
--- a/boltons/setutils.py
+++ b/boltons/setutils.py
@@ -392,10 +392,14 @@ class IndexedSet(MutableSet):
     def iter_slice(self, start, stop, step=None):
         "iterate over a slice of the set"
         iterable = self
-        if start is not None:
-            start = self._get_real_index(start)
-        if stop is not None:
-            stop = self._get_real_index(stop)
+        # islice consumes the apparent (dead-slot-free) stream, so the bounds
+        # must stay in apparent index space. Running them through
+        # _get_real_index() inflated them by the dead slots, over-counting once
+        # anything had been removed. Normalize negatives the way slicing does.
+        if start is not None and start < 0:
+            start = max(len(self) + start, 0)
+        if stop is not None and stop < 0:
+            stop = max(len(self) + stop, 0)
         if step is not None and step < 0:
             step = -step
             iterable = reversed(self)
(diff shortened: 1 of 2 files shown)


mahmoud:
Appreciate the report! When I dug in, the fix wasn't quite complete, so I knocked it out in #424. Thanks!

chuenchen309:
Thanks for finishing this off, and for keeping the reference in #424. You're right that #423 only covered the slice path — folding in the single-index `__getitem__` bounds (normalizing negatives and rejecting out-of-range against the apparent length) is the more complete fix. Glad the dead-slot analysis was useful.

PR #428, opened 2026-07-17
Fix ZeroDivisionError in backoff/backoff_iter with factor=1.0 and no count

santhreal:
backoff_iter treats factor >= 1.0 as valid (it raises only for factor < 1.0), and factor=1.0 works fine with an explicit count or count='repeat'. But when count is None the count is inferred with math.ceil(math.log(stop/denom, factor)), and math.log(x, 1.0) divides by zero, so backoff(1, 10, factor=1.0) raises ZeroDivisionError. backoff(5, 5, factor=1.0) hits the same crash even though backoff(5, 5) returns [5.0].

The count-inference branch now special-cases factor=1.0: it yields a single value when start == stop, and raises a clear ValueError otherwise (a non-growing factor cannot reach a larger stop, so the count is genuinely uninferable). Explicit-count, 'repeat', factor > 1.0, and start=0 paths are unchanged.

Repro:

    from boltons.iterutils import backoff
    backoff(1, 10, factor=1.0)   # ZeroDivisionError before, clear ValueError now
    backoff(5, 5, factor=1.0)    # ZeroDivisionError before, [5.0] now

### DIFF
--- a/boltons/iterutils.py
+++ b/boltons/iterutils.py
@@ -648,8 +648,14 @@ def backoff_iter(start, stop, count=None, factor=2.0, jitter=False):
         raise ValueError('expected stop >= start, not %r' % stop)
     if count is None:
         denom = start if start else 1
-        count = 1 + math.ceil(math.log(stop/denom, factor))
-        count = count if start else count + 1
+        if factor == 1.0:
+            if start != stop:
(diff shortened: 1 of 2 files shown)


mahmoud:
Appreciate it!

PR #429, opened 2026-07-17
Pluralize words ending in "x" with "-es" (box -> boxes)

Sreekant13:
pluralize() adds "-es" for words ending in "s", "ch", or "sh", but it was missing "x", so it returned "boxs", "foxs", and "taxs" instead of "boxes", "foxes", and "taxes". This adds "x" to that rule.

Irregular "-x" words such as "ox" are unaffected because they are resolved from the irregular map before the rule runs. I left singularize() untouched, since the reverse direction is genuinely ambiguous (axes could singularize to axe or axis).

Added test_pluralize_x; it fails on the current code.

### DIFF
--- a/boltons/strutils.py
+++ b/boltons/strutils.py
@@ -258,7 +258,7 @@ def pluralize(word):
         plural = irr_plural
     elif word.endswith('y') and word[-2:-1] not in 'aeiou':
         plural = word[:-1] + 'ies'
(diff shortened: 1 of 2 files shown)


mahmoud:
Thanxes!

Sreekant13:
Thanks for the quick review and merge, and for maintaining boltons. Happy to help.

PR #450, opened 2026-08-13
Fix Bits.__getitem__ returning False for every negative index

NotAFlightRisk:
Noticed `bits[-1]` comes back `False` whatever's actually in there.

```python
>>> from boltons.mathutils import Bits
>>> b = Bits('1011')
>>> b[3]
True
>>> b[-1]      # same bit
False
>>> b[-2:]     # slices are fine
Bits('11')
>>> b[-99]     # no IndexError either
False
```

`__getitem__` only guards `k >= self.len`, so a negative `k` puts the mask above the top of the stored int and the AND always comes out 0. The slice branch escapes it because it goes through `as_bin()`.

I've matched the shape from d0a284f on `IndexedSet._get_real_index` rather than picking my own. Shout if you'd rather negatives just raised, easy swap.

### DIFF
--- a/boltons/mathutils.py
+++ b/boltons/mathutils.py
@@ -170,7 +170,9 @@ class Bits:
         if type(k) is slice:
             return Bits(self.as_bin()[k])
         if type(k) is int:
-            if k >= self.len:
+            if k < 0:
+                k += self.len
+            if k < 0 or k >= self.len:
                 raise IndexError(k)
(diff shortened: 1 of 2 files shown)


NotAFlightRisk:
@mahmoud would you be able to take a look if you get a moment?
happy to close or change if you'd like, but it would be good to get some fix for this in

Issue #90, opened 2016-09-09
Adding a boltons.statsutils.mode() function


ajmazurie:
There are `median` and `mean` functions, but no `mode` for categorical variables.

mahmoud:
Sure! Sounds like a fine enhancement. Would you care to give it a go? Also, I'm over here wondering if a bucketized version might help make it more useful for float variables, or do you think that would make it more confusing?

ajmazurie:
I would restrict a `mode` function to discretized values (integers or strings). What you suggest sounds more like a histogram function, such as `numpy.histogram` where the user  has to say if two values (such as floats) belong to the same bucket or not, either by requesting a given number of buckets or requesting a given bucket width.

I would be more than happy to contribute a patch (never done it before, so bear with me!)

tiwo:
Python >=3.4 comes with the [statistics module](https://docs.python.org/3/library/statistics.html), providing median, mean, mode, harmonic mean (mentioned in boltons/TODO.rst), variance/standard deviation. 

Maybe this means that providing these functions in boltons isn't as urgent, or maybe it means that they can just be backported / copy-and-pasted from the standard library (I believe the Python license is compatible).

mahmoud:
Hmm, the 3.4+ stdlib version is pretty simplistic. It doesn't really encapsulate the dataset-based approach (with caching, etc.). It also inexplicably introduces the complexity of sample-based statistics by applying [Bessel's correction](https://en.wikipedia.org/wiki/Bessel%27s_correction) all over the place, when most simple applications are safe to presume population-based approaches. But, after stripping away those two design oversights, if you still find something worth cribbing, would be happy to include it! I still hear from many Python 2.6 boltons users, and they're always happy to have more features. :)

PR #453, opened 2026-08-18
Add statsutils.mode() for the most common value

ChrisJr404:
Adds a `mode` measure to `statsutils`, closing #90.

`Stats` already exposes `mean` and `median` as central-tendency measures (each with a module-level convenience function), but there was no `mode`. This fills that gap.

Implementation follows the existing pattern exactly: `mode` is a `_StatsProperty` backed by `_calc_mode`, so it caches like the other measures and automatically gets a module-level `mode()` helper via the convenience-function loop. It uses the already-imported `collections.Counter`, so it works for non-numeric, categorical data as well as numbers:

```python
>>> from boltons.statsutils import mode, Stats
>>> mode([2, 1, 3, 1])
1
>>> mode(['a', 'b', 'b', 'c', 'c', 'c'])
'c'
>>> Stats(range(5)).mode
0
```

When several values are equally common, the one appearing first in the data is returned, matching the standard library's `statistics.mode`. Empty data falls back to the object's configured `default`, consistent with the other measures.

Added a `test_mode` case and doctests (which run under `--doctest-modules`); the module docstring's measure list now mentions `mode`. Full `test_statsutils` suite and doctests pass locally.

### DIFF
--- a/boltons/statsutils.py
+++ b/boltons/statsutils.py
@@ -30,7 +30,7 @@
 
 """``statsutils`` provides tools aimed primarily at descriptive
 statistics for data analysis, such as :func:`mean` (average),
-:func:`median`, :func:`variance`, and many others,
+:func:`median`, :func:`mode`, :func:`variance`, and many others,
 
 The :class:`Stats` type provides all the main functionality of the
 ``statsutils`` module. A :class:`Stats` object wraps a given dataset,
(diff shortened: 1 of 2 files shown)


mahmoud:
Letting the agent rip eh?? Well, I don't mind, you+ solved it in the statsutils way, not a straight copy from the stdlib version. Thanks!

PR #454, opened 2026-08-18
Use the sep argument in args2sh and args2cmd

ckarnell:
Both functions take `sep`, document it, and then hardcode a space.

```python
>>> args2sh(['aa', 'bb'], sep='|')
'aa bb'
>>> args2cmd(['aa', 'bb'], sep='|')
'aa bb'
```

`args2sh` ends with `' '.join(ret_list)`. `args2cmd` builds its output incrementally and appends a literal `' '` between arguments, under a comment saying it is separating them. Both docstrings say "separated by *sep*".

Two lines, using the argument that's already there. Escaping and quoting are untouched, so only the joining changes.

Tests are `test_args2sh_sep` and `test_args2cmd_sep`, each checking the default still gives a space, a custom separator is used, and an argument needing quotes still gets quoted around the new separator. Suite goes 469 to 471, and `--doctest-modules` is 152 passing. Reverting `strutils.py` fails both new tests and nothing else.

The existing doctests all use the default, which is why this held up so long.

### DIFF
--- a/boltons/strutils.py
+++ b/boltons/strutils.py
@@ -798,7 +798,7 @@ def args2sh(args, sep=' '):
         # the string $'b is then quoted as '$'"'"'b'
         ret_list.append("'" + arg.replace("'", "'\"'\"'") + "'")
 
-    return ' '.join(ret_list)
+    return sep.join(ret_list)
 
 
 def args2cmd(args, sep=' '):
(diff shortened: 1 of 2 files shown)


ckarnell:
`escape_shell_args` is affected too, and I should have said so. It documents `sep` as "The separator used to join the escaped arguments" and forwards it to whichever of the two it picks, so it has been dropping the separator through them the whole time. Both styles come out right with this change:

```python
>>> escape_shell_args(['aa', 'bb'], sep='|')
'aa|bb'
>>> escape_shell_args(['aa', 'bb'], sep='|', style='cmd')
'aa|bb'
```

So it's three public functions, not two, with no extra change needed.

For what it's worth on provenance, this dates to 779c0bc in 2016, the commit that added all three. `args2sh` shipped with `' '.join(ret_list)` and `args2cmd` with `result.append(' ')`, so `sep` has never been honoured. It wasn't broken later.

mahmoud:
Letting the agent rip eh?? Well, I don't mind, a bug's a bug. Thanks!

PR #456, opened 2026-08-21
Make URL equality ignore query parameter order (#280)

### LINKED ISSUE #280: URL comparison is order-based for query parameters

az51826295-sys:
URL.__eq__ compared query_params directly, so two URLs carrying the same parameters in a different order compared unequal, as reported in #280.

This normalizes query params to a sorted, order-agnostic form for the equality check only - all other URL components still participate in equality as before. Includes regression tests covering reordered params, repeated keys, and the still-unequal cases. Full test suite passes locally.

Disclosure: the initial fix was produced by an AI coding agent and then human-reviewed; tests verified locally.

Closes #280

### DIFF
--- a/boltons/urlutils.py
+++ b/boltons/urlutils.py
@@ -800,9 +800,31 @@ class URL:
     def __unicode__(self):
         return self.to_text()
 
+    @staticmethod
+    def _norm_query_params(query_params):
+        "Normalize query params into an order-agnostic, comparable form."
+        if query_params is None:
+            return None
+        try:
+            items = list(query_params.items(multi=True))
+        except (AttributeError, TypeError):
+            try:
+                items = list(query_params.items())
+            except AttributeError:
+                return query_params
+        return sorted(items, key=lambda kv: (str(kv[0]), str(kv[1])))
+
     def __eq__(self, other):
         for attr in self._cmp_attrs:
-            if not getattr(self, attr) == getattr(other, attr, None):
+            self_val = getattr(self, attr)
+            other_val = getattr(other, attr, None)
+            if attr == 'query_params':
+                # query parameter order is not significant for equality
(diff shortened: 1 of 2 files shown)


mahmoud:
Thanks for taking a swing at this. Query parameter order is significant per RFC 3986, and my note in #280 still stands: a subclass with order-insensitive equality is the way to get this behavior. As written this also leaves `__hash__` out of sync with the new `__eq__`, which breaks URLs in sets and dicts. Closing.

PR #459, opened 2026-08-27
fix: emit the yotta (Y) unit in bytes2human

nikolauspschuetz:
`bytes2human` never emitted the largest unit **`Y`** (yotta): `_SIZE_RANGES` zips the size bounds with their successors, so the loop's `symbol` only ranges over `B`…`Z`, and yotta-scale values overflow into `Z` — `bytes2human(1024**8)` returned `'1024Z'` instead of `'1Y'`.

Pairs the last bound with an `(inf, '')` sentinel so the `Y` range is reachable, and adds a doctest.

Verified: `bytes2human(1024**8) == '1Y'` (and `5Y`), `Z` still correct, all `strutils` doctests pass.

<sub>Developed with AI assistance (Claude Code); I directed, reviewed, and verified it locally.</sub>


mahmoud:

PR #461, opened 2026-08-29
`ecoutils`: skip the lookups that the `scrub` parameter removes

kdeldycke:
This PR prevent the `get_profile` utility to query external resources of fields that will later get discarded if `scrub=True`.

This is speed up the method in `scrub` mode, and also prevents `os.getcwd()` to fail and `socket.getfqdn()` to performs a reverse DNS lookup.


kdeldycke:
Thanks @mahmoud for the merge! :)

mahmoud:
Sorry it took so long! Buried by these waves of extra friendly LLMs :)

PR #463, opened 2026-09-05
Honor the encoding when reading lines in reverse

codewithfourtix:
`reverse_iter_lines()` overwrites a text stream's encoding with UTF-8 and drops an explicit encoding for binary streams. Use the supplied encoding, falling back to the stream's encoding when available.

Adds Latin-1 text-stream and binary-stream regression tests.


mahmoud:

PR #466, opened 2026-09-05
Keep text streams usable after reverse iteration

codewithfourtix:
Reverse iteration detaches the buffer from a caller-owned text stream. This makes later reads and even exiting its `with` block raise `ValueError`. Read from the existing buffer without detaching it.

Adds a regression that reverses, seeks, reads, and closes the same text stream.


mahmoud:

Issue #439, opened 2026-07-29
IndexedSet: difference_update/intersection_update are quadratic for large removals


LD-RW:
## Summary

`IndexedSet.difference_update()` and `intersection_update()` are quadratic when
removing many scattered elements. Removing half of a 160,000-element set takes
1,198 ms on my machine (863x slower than builtin `set`), and 91% of that
time is spent in `IndexedSet._compact()` (flame graph attached).

## Environment

- boltons master @ e66cade
- Python 3.14.6, EndeavourOS (Arch), kernel 6.18.40-2-lts
- CPU: Intel Core i7-12700H

## Reproduction

```python
import time
from boltons.setutils import IndexedSet

for n in (10_000, 20_000, 40_000, 80_000, 160_000):
    s = IndexedSet(range(n))
    to_remove = set(range(0, n, 2))   # every other element -> scattered holes
    t0 = time.perf_counter()
    s.difference_update(to_remove)
    print(f"n={n:>7}: {(time.perf_counter() - t0) * 1000:9.1f} ms")
```

Results (best of 3):

| n | IndexedSet | builtin `set` | ratio |
|---|---|---|---|
| 10,000 | 7.8 ms | 0.08 ms | 95x |
| 20,000 | 24.6 ms | 0.16 ms | 157x |
| 40,000 | 86.1 ms | 0.33 ms | 258x |
| 80,000 | 317.3 ms | 0.61 ms | 517x |
| 160,000 | 1,197.9 ms | 1.39 ms | 863x |

Doubling n roughly quadruples the time, and the gap vs builtin `set` grows
with n instead of staying constant — quadratic scaling.

## Cause

Both methods loop `self.discard(val)` per item. Each `remove()` records a dead
interval and calls `_cull()`. Scattered removals never merge intervals, and once
the interval list exceeds the hardcoded 384 threshold, `_cull()` calls
`_compact()`, which rewrites the entire backing list — O(n). So removing k
scattered items costs about k/384 full compactions ≈ O(n·k/384); removing half
the set is O(n²/768).

cProfile confirms the mechanism: `_compact` is 91% of cumulative time with
207 calls for 80,000 removals — matching the predicted 80,000/384 ≈ 208.

The 384 cap itself seems intentional — it bounds the linear scan in
`_get_real_index()` so indexing stays fast — and that tradeoff makes sense for
single removals. But the bulk update methods know the full removal set up front,
so they can rebuild the backing list once in O(n) instead of paying repeated
compactions.

[...]

PR #470, opened 2026-09-05
Reject BarrelList item indexes past the end

codewithfourtix:
Index translation subtracts the final barrel's length even when no barrel contains the requested index. As a result, reading or deleting at `len(value)` can target an existing item. Keep the final relative index out of bounds so normal list errors apply.

Adds read, write, delete, pop, and insertion regressions.

### DIFF
--- a/boltons/listutils.py
+++ b/boltons/listutils.py
@@ -114,6 +114,8 @@ class BarrelList(list):
             if rel_idx < len_list:
                 break
             rel_idx -= len_list
+        else:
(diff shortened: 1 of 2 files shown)


mahmoud:
Superseded by #440, which predates this PR by five weeks and fixes the same `_translate_index` bound through the existing `(None, None)` sentinel, along with insert clamping and a pop empty-tail bug (benchmarks in #439). Thanks for the thorough test cases; the out-of-bounds coverage matched what landed.

PR #476, opened 2026-09-07
Honor mapping counts in ThresholdCounter.update

Jerry-val:
`ThresholdCounter.update()` documents accepting mappings of keys to integer counts, but it recognizes only the Python 2-style `iteritems()` method. On Python 3, `update({'a': 3, 'zero': 0})` counts each key once, producing `a=1`, `zero=1`, and an incorrect total. Keyword counts are affected through the same path.

Accept callable `items()` when `iteritems()` is unavailable, preserving existing legacy inputs and the per-occurrence `add()` calls that drive threshold compaction.

Add eight regression and control cases covering dict/UserDict inputs, additive counts, zero counts, keyword counts after an iterable, legacy mappings, tuple-valued keys, and equivalence to an expanded stream across compaction thresholds.

Validation: six new regression cases failed before the fix and two compatibility controls passed. `python -m pytest --doctest-modules boltons tests -q`: **633 passed** on Python 3.13.14. `git diff --check` passed.

Checked related open and closed PRs before implementation. #460 changes `most_common()` only, and #382 only adds annotations to this method; neither fixes mapping updates.

Prepared with Codex assistance and independently reviewed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ scheme (`YY.MINOR.MICRO`).
 
 _(unreleased)_
 
+- Fixed [`cacheutils.ThresholdCounter.update`][cacheutils.ThresholdCounter] ignoring counts supplied in Python 3 mappings and keyword arguments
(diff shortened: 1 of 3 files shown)


mahmoud:

PR #477, opened 2026-09-07
Fix URL serialization with a password and empty username

rzaman8677:
URL.to_text() currently drops user information when a password is present but the username is empty. For example, `URL('http://:secret@example.com/path').to_text()` becomes `http://example.com/path`.

Include user information when either a username or a password is present. The default get_authority() output continues to omit user information. Add a regression test covering serialization, explicit authority output, default authority output, and an escaped password.

Validation: the regression test failed on the original code; `python -m pytest tests/test_urlutils.py --doctest-modules boltons/urlutils.py -q` passes all 137 tests after the fix.

Prepared with OpenAI Codex assistance.

### DIFF
--- a/boltons/urlutils.py
+++ b/boltons/urlutils.py
@@ -721,7 +721,7 @@ class URL:
         """
         parts = []
         _add = parts.append
-        if self.username and with_userinfo:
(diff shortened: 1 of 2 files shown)


mahmoud:

PR #478, opened 2026-09-07
Support empty and keyword-only LRI and LRU updates

Jerry-val:
`LRI` and `LRU` currently require a positional argument for `update()`, so `cache.update()` and `cache.update(a=1)` raise `TypeError`. Passing the cache itself also returns before applying keywords, so `cache.update(cache, a=2)` silently leaves the old value unchanged.

Default the positional input to an empty iterable and skip only the self-input portion of an update. Keyword entries continue through `__setitem__`, preserving value replacement, size limits, and eviction ordering.

```python
from boltons.cacheutils import LRU

cache = LRU(max_size=2, values={'a': 1})
cache.update()             # no-op
cache.update(b=2)          # {'a': 1, 'b': 2}
cache.update(cache, a=3)   # {'a': 3, 'b': 2}
```

Tests cover both cache classes, no-op statistics and eviction order, keyword-only updates, self-input with keywords, mapping and one-shot pair inputs, keyword precedence, and explicit `None` rejection. Six regression cases fail before the fix; four compatibility controls pass before and after.

Validation on Python 3.13.14:

- `python -m pytest -q tests/test_cacheutils.py`: 28 passed.
- `python -m pytest -q --doctest-modules boltons tests`: 635 passed.
- `git diff --check`: clean.

### DIFF
--- a/boltons/cacheutils.py
+++ b/boltons/cacheutils.py
@@ -293,11 +293,11 @@ class LRI(dict):
                 self[key] = default
                 return default
 
-    def update(self, E, **F):
+    def update(self, E=(), **F):
         # E and F are throwback names to the dict() __doc__
         with self._lock:
             if E is self:
-                return
+                E = ()
             setitem = self.__setitem__
             if callable(getattr(E, 'keys', None)):
                 for k in E.keys():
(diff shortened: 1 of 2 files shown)


mahmoud:

PR #512, opened 2026-09-20
Correct MultiFileReader read and rewind behavior

rupayon123:
## The change

Correct MultiFileReader cursor behavior: read(0) no longer consumes all input, negative-size reads return the remaining data, and seek(0) resets which underlying file is active. Previously, rewinding after crossing a file boundary could silently omit earlier files.

Eight regression cases fail before the fix. All 682 tests and module doctests pass afterward on Python3.11.15, with bytes and text sources and rewind before/after EOF covered.

## The backstory

This follows the utility-library investigation in #502. Concatenated file readers should keep the same cursor semantics as their source streams. This was found through source inspection and controlled regression tests.

## AI assistance

- Harness/tooling: OpenAI Codex acting on behalf of rupayon123.
- Model: GPT-6-based Codex; exact backend variant not exposed.
- Original task excerpt: “Execute my daily GitHub contribution and maintenance workflow on my behalf. Do real, verified work, not only research, a plan, configuration changes, or a status summary.” Required reproducible, tested fixes and no duplicate patches.


Voyagerroc-Lab (approved):
Reviewed d23a2b8. The cursor changes keep `read(0)` non-consuming, make negative-size reads return the remaining data, and restore the active file when rewinding after a boundary or EOF.

On Windows / CPython 3.14.7, **95 tests/doctests passed** across `tests/test_ioutils.py`, `tests/test_multifile_state.py`, and `boltons/ioutils.py`. An additional 30 deterministic operation sequences matched `StringIO`/`BytesIO` over concatenated inputs, including empty component files, zero/negative reads, repeated EOF reads, and repeated rewinds; the parent implementation disagreed in 16 of those sequences. This validation was scoped to the affected I/O code, not the full repository suite.

Review and validation performed with OpenAI Codex at the account owner's request.

PR #513, opened 2026-09-21
Fix spooled stream write compatibility and backwards text seeks

rupayon123:
## The change

Make spooled streams return their write count, and let SpooledBytesIO accept buffer-protocol inputs. Wrapping the binary stream in `io.BufferedWriter` currently fails on flush because it supplies a memoryview; successful writes also return None instead of their byte or character count. The rollover calculation now uses the buffer size in bytes, including multi-byte memoryviews.

A second fix corrects backwards character-relative seeks in SpooledStringIO and rejects negative absolute destinations without moving the stream. Previously a backwards seek consumed the rest of the buffer while reporting the requested earlier position. UTF-8 navigation now restarts from the beginning for backwards movement.

Six write regression cases and six seek regression cases fail before their respective fixes. All 685 combined tests and module doctests pass afterward on Python 3.11.15. Coverage includes rolled and in-memory streams, Unicode character counts, BufferedWriter flushes, and rollover with a multi-byte memoryview. Documentation examples are updated for the returned values.

## The backstory

This follows the utility-library investigation in #502. The bug was found through source inspection and a real BufferedWriter integration regression, rather than a production incident. The change keeps spooled streams compatible with standard I/O callers.

## AI assistance

- Harness/tooling: OpenAI Codex acting on behalf of rupayon123.
- Model: GPT-6-based Codex; exact backend variant not exposed.
- Original task excerpt: “Execute my daily GitHub contribution and maintenance workflow on my behalf. Do real, verified work, not only research, a plan, configuration changes, or a status summary.” Required reproducible, tested fixes and no duplicate patches.


induane (changes requested):
I can't imagine I have actual approval ability for this repo, but since I contributed the original SpooledStringIO and SpooledBytesIO implementations, I suppose I at least do have opinions ;-) 

In any case this looks quite good,  my only real complaint is the way the tests are implemented - I don't know that we need two new files for two new categories of tests on an existing module. 

You should probably more closely follow existing code standards and place the tests in the test_ioutils.py file.

rupayon123:
I moved the spooled stream regression cases into the existing `tests/test_ioutils.py`, using the module's `TestCase` style, and removed the two new test files. This is in d68651ff9e862a4ebf9dd8c3c6c286ab544d59d2; the production implementation is unchanged.

I reran the full tests and module doctests on Python 3.11.15: 678 passed, with 12 passing subtests covering the rolled/in-memory cases. `git diff --check` also passes. Remote CI on the new head is still pending maintainer approval, so I am leaving that result open.

induane:
LGTM - I don't know if the maintainers would quibble over the lack of docstring descriptions in the tests or not but that's the only possible nit I'd throw out.

induane (approved):

mahmoud:
If @induane approves it, that's all I need! (Also hey induane! Pleasantly surprised to see your name, and the new profile pic!)

rupayon123:
Thanks for reviewing and merging this, both of you. Moving the cases into the existing test module made the change easier to follow.

induane:
@mahmoud good to hear from you too! Hit me up sometime via email or whatever. 

I changed my profile pic because I was afraid it was too cool for recruiters who view my github :laughing:

